PR #1, opened 2025-11-04
Ortac

katrinafyi:


agle:
I want to merge this but not with the example test that doesn't test anything substantial, good candidates are

- the page table structure used by the interpreter
- the fresh identifier system, but its contract is ugly
- would value domain soundness make sense in ortac or just properties?

katrinafyi:
superseded by https://github.com/agle/bincaml/pull/8 because ortac is too hard and opaque, so we should just use qcheck-stm directly.

PR #8, opened 2025-12-19
qcheck-stm

katrinafyi:
BIG: rename Util to Bincaml_util

adds qcheck-stm and uses it to test that the PageTable module is functionally equivalent to a map of Z.t to char. 

to run the test, you can run the associated .exe. this shows an example error which would happen if you changed read to always return a bv1. the message contains the executed commands and the _actual_ output of the SUT. sadly, the _expected_ output from the abstract model is not shown.

this is because the abstract model doesn't have a notion of "results" arising from commands, it merely updates its own abstract state. 
```console
$ dune exec test/lang/pagetable_qcheck.exe
random seed: 162710552
generated error fail pass / total     time test name
[✗]    1    0    1    0 / 1000     0.0s STM sequential tests

--- Failure --------------------------------------------------------------------

Test STM sequential tests failed (4 shrink steps):

   Read {addr = 8223555767233468559; nbytes = 1}


+++ Messages ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

Messages for test STM sequential tests:

  Results incompatible with model

   Read {addr = 8223555767233468559; nbytes = 1} : 0x0:bv1
================================================================================
failure (1 tests failed, 0 tests errored, ran 1 tests)
```


test/lang/pagetable_spec.ml
```diff
@@ -0,0 +1,144 @@
...
+  let postcond cmd oldst res =
+    match (cmd, res) with
+    | Read { addr; nbytes }, STM.Res ((BitvecTy, _), bv) ->
+        bv.w = nbytes * 8
+        && Bitvec.to_bytes bv |> Bytes.to_seq
+           |> Seq.fold_lefti
+                (fun ok i c -> ok && read_byte ~addr:Z.(addr + ~$i) oldst = c)
+                true
+    | Write _, STM.Res ((STM.Unit, _), ()) -> true
+    | _ -> false
```
agle: I feel like it would make more sense to just check the entire abstract state (the domain st.bindings \cup [addr, addr + i]). I get that addr gen is set up so its probable the addresses but this version obscures whether the error was caused by a read or a write operation, but I don't know if it would make an observable difference to test performance.

lib/contributing.mld
```diff
@@ -6,11 +6,11 @@ This guide explains who to get started as a contributor.
...
 
-All files should import the [util.common] library which contains extensions to
+All files should import the [Bincaml_util.Common] library which contains extensions to
 the OCaml standard library in the [containers] library, this is done by
 depending on the [bincaml.util] library and putting at the top of the file.
 
-[open Util.Common]
+[open Bincaml_util.Common]
 
 Use the documentation for [containers] as a reference before the standard
```
agle: thanks for updating this, we could probably do with some doc when qcheck-stm is the appropriate choice and why we use qcheck-stm over ortac or plain qcheck or unit test.

agle (approved):
Thanks, comments are optional

test/lang/pagetable_spec.ml
katrinafyi: I guess it is obscured, but it's the same as ortac's qcheck-stm backend. Checking the abstract state means that you'd need (something like) a projection function from concrete to abstract. I wanted this test to be as minimal as possible, so that it can be directly compared to ortac's qcheck-stm backend.

PR #9, opened 2025-12-19
def/use analysis

agle:
- compute def-use graph from SSA form and run analyses over this --- uses patricia tree to reuse the ocamlgraph chaotic iteration without too much inefficiency, should write custom solver to fully benefit though (use one mutable state, like Fix.Dataflow with widening?) 
- functorised abstractions of value/state domains
- factor out recursion scheme
- an example value analysis for boolean value; probably makes sense on regular IR as well as SSA or def-use

to do

- [x] fork valuegraph thing into a different branch as its unrelated
- [x] some form of testing (could do with more obviously)

next work

- [ ] evaluation of analyses wrt performance
- [ ] framework for printing each eval step in the local analysis framework as a trace
- [ ] some simple-ish setup for printing the analysis result as a graph
- [ ] implement interpreter-based sva soundness litmus property test
- [ ] considerations with culling edges we find are dead from the analysis ; lift state lattice to add bot value
- [ ] write a transform that uses the analysis (e.g. just replace guards/asserts with true/false when known)
- [ ] write a tutorial based on this analysis and transform

### DIFF
--- a/lib/analysis/intra_analysis.ml
+++ b/lib/analysis/intra_analysis.ml
@@ -5,7 +5,7 @@ open Common
 open Containers
 open Lattice_types
 
-module EvalValueAbstraction (V : ValueAbstraction) = struct
+module EvalExpr (V : ValueAbstraction) = struct
   type t
 
   let eval read expr =
@@ -19,24 +19,34 @@ module EvalValueAbstraction (V : ValueAbstraction) = struct
       | ApplyIntrin (op, es) -> V.eval_intrin op es
       | _ -> failwith "unsupported"
     in
-    Lang.Expr.BasilExpr.cata eval_alg expr
+    V.E.cata eval_alg expr
+end
+
+module EvalValueAbstraction
+    (V : ValueAbstraction with module E = Expr.BasilExpr) =
+struct
+  type t
+
+  module Eval = EvalExpr (V)
+
+  let eval read expr = Eval.eval read expr
 end
 
 module EvalStmt
     (V : ValueAbstraction)
-    (S : StateAbstraction with type key_t = Var.t with type val_t = V.t) =
+    (S : StateAbstraction with type val_t = V.t with type key_t = V.E.var) =
 struct
   type t
 
-  module EV = EvalValueAbstraction (V)
+  module EV = EvalExpr (V)
 
-  let stmt_eval_fwd (stmt : Program.stmt) dom =
+  let stmt_eval_fwd stmt dom =
     Stmt.map ~f_lvar:id
       ~f_rvar:(fun v -> S.read v dom)
       ~f_expr:(EV.eval (fun v -> S.read v dom))
       stmt
 
-  let stmt_eval_rev (stmt : Program.stmt) dom =
+  let stmt_eval_rev stmt dom =
     Stmt.map ~f_lvar:(fun v -> S.read v dom) ~f_rvar:id ~f_expr:id stmt
 end
 
--- a/lib/analysis/defuse_bool.ml
+++ b/lib/analysis/defuse_bool.ml
@@ -113,8 +113,15 @@ end
 
 module Domain = Intra_analysis.MapState (IsZeroLattice)
 
+module Eval =
+  Intra_analysis.EvalStmt
+    (struct
+      include IsZeroValueAbstraction
+      module E = Lang.Expr.BasilExpr
+    end)
+    (Domain)
+
 module TransferFunc = struct
-  module Eval = Intra_analysis.EvalStmt (IsZeroValueAbstraction) (Domain)
   open IsZeroLattice
   include Domain
 
--- a/lib/analysis/lattice_types.ml
+++ b/lib/analysis/lattice_types.ml
@@ -18,13 +18,14 @@ end
 (** an abstract value providing opertions of basil ir *)
 module type ValueAbstraction = sig
   include Lattice
+  module E : Expr.ExprType
 
   (** evaluate operators *)
 
-  val eval_const : Lang.Ops.AllOps.const -> t
-  val eval_unop : Lang.Ops.AllOps.unary -> t -> t
-  val eval_binop : Lang.Ops.AllOps.binary -> t -> t -> t
-  val eval_intrin : Lang.Ops.AllOps.intrin -> t list -> t
+  val eval_const : E.const -> t
+  val eval_unop : E.unary -> t -> t
+  val eval_binop : E.binary -> t -> t -> t
+  val eval_intrin : E.intrin -> t list -> t
 end
 
 (** a value abstraction providing variable store/load operations *)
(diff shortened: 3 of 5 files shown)


lib/analysis/defuse.ml
```diff
@@ -0,0 +1,212 @@
...
+  module DuAnalysis = Graph.ChaoticIteration.Make (G) (State)
+
+  let analyse root ?(init = fun v -> State.analyze_vert v (StateDomain.bottom)) ~widen_set ~delay_widen
+      p =
+    let g = def_use_graph p in
+    DuAnalysis.recurse g (Topo.recursive_scc g root) init widen_set delay_widen
+
+  let analyse_graph g root ?(init = fun v -> StateDomain.bottom) ~widen_set
+      ~delay_widen p =
+    DuAnalysis.recurse g (Topo.recursive_scc g root) init widen_set delay_widen
```
katrinafyi: So... does `def_use_graph` or `DuAnalysis.recurse` actually do the analysis?

lib/analysis/defuse.ml
```diff
@@ -0,0 +1,212 @@
...
+    | Stmt s -> Hash.combine2 251 (Hashtbl.hash s)
+    | Phi (l, r) -> Hash.combine2 271 (Hash.list Var.hash r)
+    | o -> Hashtbl.hash o
+end
+
+module UseDef = Graph.Persistent.Digraph.ConcreteBidirectional (Vertex)
+module B = Graph.Builder.P (UseDef)
+module MDeps = CCMultiMap.Make (Var) (Vertex)
+
+type defuse = { var_to_use : MDeps.t; var_to_def : MDeps.t }
```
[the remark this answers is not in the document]
agle: The goal of this file is to build a SSA form data dependency graph (the def-use graph) and then provide a framework for defining a dataflow analysis over that graph. Because of SSA the  analysis that goes into building this graph is pretty trivial every statement with a rhs variable v gets an edge to every statement with an lhs variable v. This graph is bidirectional in order to perform both a backwards or forwards dataflow analysis. Using Graph.ChaoticIteration. To this end it doesn't do the location bookeeping neccessary to do def/use lookups on the original graph.

lib/analysis/defuse.ml
agle: `recurse` is the entypoint to ocamlgraph's chaoticiteration _data flow_ solver, `defuse_graph` builds  the graph this solver solves over.

lib/analysis/defuse.ml
```diff
@@ -0,0 +1,212 @@
...
+      ~delay_widen p =
+    DuAnalysis.recurse g (Topo.recursive_scc g root) init widen_set delay_widen
+end
+
+module AnalysisFwd
+    (V : Intra_analysis.ValueAbstraction)
+    (TRF : Intra_analysis.ForwardStmtTransfer with type t = V.t) =
+struct
+  module TF = Intra_analysis.StateTransferFwd (V) (TRF)
+  include DefUseGraphAnalysis (UseDef) (V) (TF)
```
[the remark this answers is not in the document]
agle: `AnalysisFwd` and `AnalysisRev` have different functor signatures: note `ForwardStmtTransfer` vs `ReverseStmtTransfer` and the corresponding `StateTransferFwd` and `StateTransferRev`, this is because forwards and backwards abstract interpretation has different meaning and type signatures (e.g. whether abstract values appear on the RHS or the LHS of the statement to the statement transfer function). I've gotten rid of the include here as its unneccessary.

lib/analysis/intra_analysis.ml
```diff
@@ -74,6 +103,80 @@ module type Analysis = sig
...
+  val transfer : Program.stmt -> t -> t
+end
+
+module type ForwardStmtTransfer = sig
+  type t
+
+  val do_assume :
+    (Var.t -> t) -> Program.e -> is_guard:bool -> (Var.t * t) Iter.t
+
+  val do_assert : (Var.t -> t) -> Program.e -> (Var.t * t) Iter.t
```
katrinafyi: StateTransferFwd doesnt seem to use these, only StateTransferRev does. ig it should be removed from the module type if it's not used? but why do the two directions differ in this way?

lib/analysis/intra_analysis.ml
```diff
@@ -74,6 +103,80 @@ module type Analysis = sig
   val tf_stmt : t -> Program.stmt -> t
 end
 
+module type Transfer = sig
+  type t
+
+  val transfer : Program.stmt -> t -> t
+end
+
+module type ForwardStmtTransfer = sig
```
katrinafyi: what's the difference between Transfer and ForwardStmtTransfer, when Transfer.transfer also takes a Program.stmt?

so ForwardStmtTransfer gets passed to StateTransferFwd which gets passed to DataflowAnalysis where it gets built into a StateDomain.

i guess ForwardStmtTransfer is some higher level representation of a transfer function? this should be made clear in the doc comments so the people know what interface they should try to satisfy.

agle: ForwardStmtTransfer is a refinement of Transfer for a forwards flow analysis using a lattice which maps variables to abstract values. Transfer is used in the solver's context where it doesn't care what the transfer operations are precisely and only needs the lattice operations.

lib/analysis/dataflow_graph.ml
```diff
@@ -0,0 +1,307 @@
...
+end
+end
+
+(** forwards dataflow analysis over dfg *)
+module AnalysisFwd
+    (V : Intra_analysis.ValueAbstraction)
+    (TRF : Intra_analysis.ForwardStmtTransfer with type t = V.t) =
+struct
+  module TF = Intra_analysis.StateTransferFwd (V) (TRF)
+  module A = DataflowAnalysis(DFGraph) (V) (TF)
```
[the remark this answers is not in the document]
agle: I'm not sure what you mean, abstract value and transfer function seem like the sensible parameters to a dataflow analysis. Requiring the user to construct a DataFlowAnalysis would leak more of the implementation (i.e. them to know whether or not to reverse the view of the graph). Maybe DataflowAnalysis could drop the V parameter by putting it in TF?

lib/analysis/intra_analysis.ml
agle: Forward transfer should use them too

lib/analysis/intra_analysis.ml
```diff
@@ -74,6 +103,80 @@ module type Analysis = sig
...
+struct
+  open struct
+    module StateDomain = MapState (V)
+    module EV = EvalValueAbstraction (V)
+  end
+
+  type t = StateDomain.t
+
+  let do_assume read e ~is_guard = SF.do_assume read e ~is_guard
+  let do_assert read e = SF.do_assert read e
```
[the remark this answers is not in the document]
agle: IDK wym but I think these should just be deleted, SF.do_assume is called directly by `transfer` which is all thats needed (and forwards should do the same).

PR #13, opened 2026-01-12
Soundness check

katrinafyi:


katrinafyi:
Abandoning because outdated. That's not to say the idea is bad or we shouldn't do it :)

Linked to https://github.com/agle/bincaml/issues/74

PR #14, opened 2026-01-14
Simple Type Check

JTrenerry:
Adds a new ProcCheck to ensure type correctness of an IR program.
- [x] Change String.append to Printf.sprintf
- [ ] Descriptive error messages
- [ ] Write a test suite checking invalid types

### DIFF
--- a/lib/transforms/type_check.ml
+++ b/lib/transforms/type_check.ml
@@ -3,14 +3,14 @@
 open Bincaml_util.Common
 open Lang
 open Expr
+open Printf
 
 type type_error = StatementEqualityError of { text : string }
 
 let print_errors ls =
   List.iter
     (fun msg ->
-      match msg with
-      | StatementEqualityError { text = t } -> Printf.printf "%s\n" t)
+      match msg with StatementEqualityError { text = t } -> printf "%s\n" t)
     ls
 
 let binary_same_types arg1 arg2 (typ : Types.t) =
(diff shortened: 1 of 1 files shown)


lib/transforms/type_check.ml
```diff
@@ -0,0 +1,323 @@
...
+      *)
+      List.concat
+        (List.fold_left
+           (fun acc arg -> type_check arg :: acc)
+           [
+             List.append
+               (compare_maps BasilExpr.type_of args real_args stmt)
+               (compare_maps Var.typ lhs output stmt);
+           ]
+           (List.map (fun (_, v) -> v) (StringMap.to_list args)))
```
agle: You could write this  more cleanly like below; `flat_map` is combines the map and concat operations.

```ocaml
let params_check = List.append
               (compare_maps BasilExpr.type_of args real_args stmt)
               (compare_maps Var.typ lhs output stmt) in
let args =  StringMap.values args |> Iter.to_list |> List.flat_map type_check in
List.append params_check args
```

agle (commented):
The diff view here looks really odd, maybe like its rebased main onto this branch rather than this branch onto main? would you mind messaging me what you ran to rebase it?

JTrenerry:
> The diff view here looks really odd, maybe like its rebased main onto this branch rather than this branch onto main? would you mind messaging me what you ran to rebase it?

Yeah, I think I did the reverse thinking it was like merge in the direction. I was on the branch and did `git rebase main` (I believe).

PR #21, opened 2026-01-22
IDE solver

b-paul:
Rewrites the IDE solver. Interproc livevar analysis has been implemented and tested to some extent.


lib/analysis/ide.ml
```diff
@@ -0,0 +1,766 @@
...
+      let (x : Loc.t * DL.t * DL.t) =
+        Q.extract worklist |> Option.get_exn_or "queue empty"
+      in
+      let l, d1, d2 = x in
+      let ost = get_summary l in
+      let e1 = dldlget d1 d2 ost in
+      IDEGraph.G.succ_e graph l |> Iter.of_list
+      |> Iter.iter (fun e ->
+          let from, target = match e with from, _, target -> (from, target) in
+          match IDEGraph.G.E.label e with
```
agle: Can we factor this match into a function?

lib/analysis/ide.ml
```diff
@@ -0,0 +1,766 @@
...
+                    Hashtbl.get entry_to_exit_cache (callinfo.callee, d3)
+                    |> Option.map (fun m ->
+                        DlMap.to_iter m
+                        |> Iter.iter (fun (d4, e3) ->
+                            let e321 = D.compose e3 e21 in
+                            D.transfer_return callinfo.ret d4
+                            |> Iter.map (fun (d5, e4) ->
+                                ((d1, d5), D.compose e4 e321))
+                            |> propagate worklist summaries priority
+                                 (get_summary aftercall) aftercall))
```
agle: Lots of places you could probably flatten the indentation by using iter and flat map rather then nesting into the fold/map function argument, eg `a |> Option.to_iter |> Iter.flat_map DLMap.to_iter |> ...`  because I read left to right lol

lib/analysis/ide.ml
```diff
@@ -0,0 +1,773 @@
...
+        DlMap.to_iter m |> Iter.map (fun x -> (d1, x)))
+    |> Iter.to_string ~sep:", " (fun (v, (v', i)) ->
+        "(" ^ DL.show v ^ "," ^ DL.show v' ^ "->" ^ D.show i ^ ")")
+
+  let empty_summary = DlMap.empty
+
+  type analysis_state = D.Value.t DataMap.t [@@deriving eq, ord]
+
+  module Worklist (D : Heap.TOTAL_ORD) = struct
+    module PQ = Heap.Make_from_compare (D)
```
agle: There's containers-data MutHeap, it may perform differently I don't know.

lib/analysis/ide.ml
```diff
@@ -0,0 +1,773 @@
...
+
+  module P1K = struct
+    type t = Loc.t * DL.t * DL.t
+
+    let compare (a, b, c) (d, e, f) =
+      Pair.compare
+        (Pair.compare Loc.compare DL.compare)
+        DL.compare
+        ((a, b), c)
+        ((d, e), f)
```
agle: `Ord.triple Loc.compare DL.compare DL.compare`

lib/analysis/ide_live.ml
```diff
@@ -0,0 +1,162 @@
...
+    let open Stmt in
+    match d with
+    | Lambda -> (
+        match stmt with
+        | Instr_Assign _ -> Iter.singleton (d, IdEdge)
+        | _ ->
+            Stmt.free_vars_iter stmt
+            |> Iter.fold
+                 (fun i v -> Iter.cons (Label v, ConstEdge live) i)
+                 (Iter.singleton (d, IdEdge)))
```
agle: `Iter.cons` is inefficient as it allocates thunks, this is just an  Iter.map (v -> Label v , ConstEdge live); reversing is not important? The lists are small here though so it probably doesn't matter.

agle (approved):
This is pretty close now thanks

lib/analysis/ide.ml
b-paul: It looks like it requires assigning each worklist value an integer for priority, which I don't know how to do :(

PR #22, opened 2026-01-30
Wrapping Intervals

HaydenAtFoxHaven:
Interval analysis, but on a circle!

### DIFF
--- a/lib/analysis/wrapped_intervals.ml
+++ b/lib/analysis/wrapped_intervals.ml
@@ -862,6 +862,7 @@ module WrappedIntervalsValueAbstraction = struct
       | `ZeroExtend k -> zero_extend a k
       | `SignExtend k -> sign_extend a k
       | `Extract (hi, lo) -> extract ~hi ~lo a
+      | `BOOLTOBV1 -> { w = Some 1; v = Top }
       | _ -> infer a top |> snd
 
   let eval_binop (op : Lang.Ops.AllOps.binary) a b =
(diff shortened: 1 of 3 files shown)


lib/analysis/wrapping_intervals.ml
```diff
@@ -0,0 +1,867 @@
...
+    {
+      w = Some (Bitvec.size lower);
+      v =
+        (if Bitvec.(equal lower (add upper (of_int ~size:(size upper) 1))) then
+           Top
+         else Interval { lower; upper });
+    }
+
+  let infer { w = w1; v = a } { w = w2; v = b } =
+    match (w1, w2) with
```
agle: I would write this like,

```ocaml
let infer a b = 
  let w = match (a.w,b.w) with
      | Some a, Some b -> assert a = b ; a
      | Some a , None -> a
      | None, Some a -> a
      | None, None -> None in
({a with w}, {b with w})

```

lib/analysis/wrapped_intervals.ml
```diff
@@ -0,0 +1,866 @@
...
+    in
+    infer s
+    @@ lub
+         (List.concat_map
+            (fun a -> List.concat_map (fun b -> rusmul a b) (cut t))
+            (cut s))
+    |> snd
+
+  (* Division implementation derived from Crab *)
+  let trim_zeroes t =
```
agle: It would be good to link the actual implementation here, I think you dropped it in zulip, [somewhere in here](https://github.com/seahorn/crab/blob/dev/include/crab/domains/wrapped_interval_impl.hpp#L224)?

PR #23, opened 2026-01-30
Known Bits Domain 

sneha-zazen:
This domain of tnums (tristate numbers) is used to reason about the bitwise uncertainty in program values.

### DIFF
--- a/docs/dev/wrapped-intervals.pdf
+++ b/docs/dev/wrapped-intervals.pdf
@@ -1,0 +1,4664 @@
+%PDF-1.7
+%€€€€
+
+1 0 obj
+<<
+  /Type /Pages
+  /Count 5
+  /Kids [306 0 R 310 0 R 313 0 R 317 0 R 324 0 R]
+>>
+endobj
+
+2 0 obj
+<<
+  /Type /Outlines
+  /First 3 0 R
+  /Last 15 0 R
+  /Count 6
+>>
+endobj
+
+3 0 obj
+<<
+  /Parent 2 0 R
+  /Next 4 0 R
+  /Title (1 Background)
+  /Dest 268 0 R
+>>
+endobj
+
+4 0 obj
+<<
+  /Parent 2 0 R
+  /Next 6 0 R
+  /Prev 3 0 R
+  /First 5 0 R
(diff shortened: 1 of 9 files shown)


lib/analysis/known_bits.ml
```diff
@@ -0,0 +1,267 @@
...
+    bind1 (fun (v, m) ->
+        tnum
+          (Bitvec.zero_extend ~extension:k v)
+          (Bitvec.zero_extend ~extension:k m))
+
+  let sign_extend k =
+    bind1 (fun (v, m) ->
+        tnum
+          (Bitvec.sign_extend ~extension:k v)
+          (Bitvec.zero_extend ~extension:k m))
```
agle: There's a lot of local opens of `Bitvec` its probably worth just opening in the module with `open Bitvec` if it doesnt conflict too much

lib/analysis/known_bits.ml
```diff
@@ -0,0 +1,267 @@
+open Bincaml_util.Common
+
+module IsKnownLattice = struct
+  let name = "tnum"
```
agle: Can you link the paper for reference

agle (commented):
Good work, its probably a good idea to also check in your documentation pdf under /docs as well

lib/analysis/known_bits.ml
```diff
@@ -0,0 +1,267 @@
...
+    bind2 (fun (av, am) (bv, bm) ->
+        if Bitvec.is_nonzero bm then Top
+        else tnum (Bitvec.lshr av bv) (Bitvec.shl am bv))
+
+  let bitASHR =
+    bind2 (fun (av, am) (bv, bm) ->
+        if Bitvec.is_nonzero bm then Top
+        else tnum (Bitvec.ashr av bv) (Bitvec.shl am bv))
+
+  let mul =
```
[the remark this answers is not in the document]
sneha-zazen: the algorithm was tweaked to be written in ocaml so it has recursive calls and cases that follows final version of mul as given in the paper but it takes inspiration from other implementations as well 
i have explained it more in the comments but lmk if i should change anything there

PR #46, opened 2026-02-16
Use compare for soundness check rather than join

HaydenAtFoxHaven:
Currently the prop tests for value soundness take the join of the abstract and concrete evaluation, then check for equality. However, this only holds true for analyses where join does not contain a reduction step.

This has been replaced with `V.compare concrete abstract <= 0` which is more accurate and holds for sound reductions.

Also the compare function for `IsZeroLattice` has been rewritten manually as the derived version caused `Top < NonZero`.

### DIFF
--- a/lib/analysis/defuse_bool.ml
+++ b/lib/analysis/defuse_bool.ml
@@ -7,10 +7,17 @@ module IsZeroLattice = struct
   let name = "isZero"
 
   type t = Top | Zero | NonZero | Bot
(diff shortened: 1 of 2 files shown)


agle:
I'll merge this for now so its easier to write tests against it until we fix it

HaydenAtFoxHaven:
I'll make a version with the less than operator in `Lattice` in another branch and migrate things to it. I'm assuming `[@@deriving ord]` is sufficient for collection stuff?

PR #50, opened 2026-02-17
Possible fix for CI

JTrenerry:
Issue seems to be the new update for containers 2 days ago, not sure why this flips the order though.

### DIFF
--- a/bincaml.opam
+++ b/bincaml.opam
@@ -18,7 +18,7 @@ depends: [
   "fix"
   "trace"
   "trace-tef"
-  "containers"
+  "containers" {<= "3.17"}
   "iter"
   "containers-data"
   "ppx_deriving"


JTrenerry:
lol wrong way

b-paul:
it looks like this commit is what broke things https://github.com/c-cube/ocaml-containers/commit/571f9f3793ea790a5dfc6df26a6f73ed4839ac48 (did a bisect with pinning versions of containers)

the only change in this commit that differs between ocaml 5.3 and ocaml 4.14.x is the implementation of `Map.of_list` and `Map.to_list`, and `of_list` is used in `loadir.ml`....... suspicious

agle:
Thanks, b-paul, it looks like the fix is to use `bindings` which defines the order, instead of `to_list` which preserves stdlib behvaiour (which is undefined for 4.*).

Issue #54, opened 2026-02-19
Nix - opam-nix


JTrenerry:
The flake has already become tiresome to maintain alongside the existing dune-project, a better alternative may be [dune-nix](https://github.com/o1-labs/dune-nix).

This should do two things (though the first is probably possible with opam-nix).

1) Allow us to use one file for both the flake opam packages and the dune packages

 > This uses a opam.export file which is generated via the opam switch. There may be a problem there though as the current flake does not provide a way to run opam commands, so if someone using the flake was to add a package via the dune-project, the only way to add it to the flake would be via manually editing the opam.export (unsure how readable this file it at the moment)

2) Allow for better build times (opam-nix is apparently very dumb and just rebuilds everything constantly)

For now, I will just make a PR keeping the flake.nix up to date from the recent changes in #52

JTrenerry:
Dune-nix seems to be a HUGE pain, there is zero docs on it and the example they give is like 10 files large. (There is a smaller one but I don't understand it cause it is partial and overlays lol)

I think the first issue can be solved without dune-nix still though.

katrinafyi:
My suggestion would be to use the nixpkgs infrastructure. It looks like you can enable fp and flambda with overrides: https://github.com/NixOS/nixpkgs/blob/8c809a146a140c5c8806f13399592dbcb1bb5dc4/pkgs/development/compilers/ocaml/generic.nix#L33. I think this would be faster and more robust, but it comes with the usual limitations of nixpkgs (that there's only one version of most packages and you have to manually package any unpackaged dependencies).

PR #55, opened 2026-02-19
Bump flake

JTrenerry:
Changes in #52 (saw the PR but completely forgot oops) will need to be changed in the flake as well.

This is annoying to keep maintained for people who don't use the flake so #54 is proposed as a solution and a better alternative to the current setup.


katrinafyi (approved):
yeah... nix :")

PR #56, opened 2026-02-19
Set and map with top

b-paul:
Adds a set type and map type with an identified top value (assuming values of the map have a top value). Some proptests are included (taken from basil).

### DIFF
--- a/lib/analysis/lattice_collections.ml
+++ b/lib/analysis/lattice_collections.ml
@@ -75,7 +75,7 @@ module LatticeMap (K : MapKey) (V : TopLattice) = struct
       type key_t = K.t
       type t = BotMap of V.t KM.t | TopMap of V.t KM.t
 
-      let name = V.name ^ "maplattice"
+      let name = V.name ^ "MapLattice"
       let bottom = BotMap KM.empty
       let top = TopMap KM.empty
 


lib/analysis/lattice_collections.ml
```diff
@@ -0,0 +1,193 @@
...
+  let name = T.name ^ "LatticeSet"
+
+  let show = function
+    | Fin s -> TSet.to_string ~start:"{" ~stop:"}" T.show s
+    | Top -> "Top"
+
+  let pretty = function
+    | Fin s ->
+        Containers_pp.(
+          text "Fin" ^ fill (text ",") (TSet.to_list s |> List.map T.pretty))
```
agle: If you make the separator `text "," ^ newline` it should wrap long lines.

lib/analysis/lattice_collections.ml
```diff
@@ -0,0 +1,193 @@
...
+
+  type t = Fin of TSet.t | Top [@@deriving eq, ord]
+
+  module T = T
+
+  let name = T.name ^ "LatticeSet"
+
+  let show = function
+    | Fin s -> TSet.to_string ~start:"{" ~stop:"}" T.show s
+    | Top -> "Top"
```
agle: This can use https://agle.github.io/bincaml/bincaml/bincaml.util/Bincaml_util/Unicode/index.html#val-top_char

lib/analysis/lattice_collections.ml
```diff
@@ -0,0 +1,193 @@
...
+        in
+        s
+        ^ (Iter.from_iter (fun f -> KM.iter (fun k v -> f (k, v)) m)
+          |> Iter.to_string ~sep:", " (fun (k, v) ->
+              Printf.sprintf "%s->%s" (K.show k) (V.show v)))
+
+      let pretty a =
+        Containers_pp.(
+          let m, s =
+            match a with BotMap m -> (m, "") | TopMap m -> (m, "TopMap ")
```
agle: I wonder whether we just write this as `(k -> v, ... , _ -> T)` or `(k -> v, ... , _ -> ⊥)`

PR #58, opened 2026-02-23
Type Inference

JTrenerry:
# Progress
- [x] Type inferencing
  - [x] Constraint generation (Relies on SVA)
  - [X] Type coalescing
  - [X] Type Automata
  - [X] Type lowering
    - [X]  Remove Ep
    - [x] Merge Nodes (old code in git so should be simple)
    - ~~mini~~ Unsure if this does anything of value
  - [x] Transform
    - [x] Add in new expressions
    - [X] Type transform
      - [x] type decls
      - [X] Exprs
      - [X] stmts
      - [x] function in / out
      ~~attrib~~
  
```sexp
(load-il "examples/irreducible_loop_1.il")
(run-transforms "ssa")
(run-transforms "cf-expressions")
(run-transforms "simplify")
(run-transforms "cf-expressions")
(dump-il "before.il")
(run-transforms "type-check")
(run-transforms "type-inference")
(run-transforms "type-check")
(dump-il "after.il")
```
  
Note: SVA does not net very good results in most cases

I ticked everything off mainly to feel good, it does not work amazingly atm
# Testing
The current testing is just, type inference loops and does it type check and look the same as the last time I type inferred.

There are some functions that have test cases, but I want to add some more to this.

Since analyse is in its own function, I should be able to do cram tests without needing diff, and possibly using leq instead of equal, but that might not be amazing


JTrenerry:
I think it would be best if this is merged once fully complete as it would allow for more complete testing.

If SVA is needed elsewhere I can cherry pick those commits, but I also would need some better testing for SVA as well.

JTrenerry:
store constraint seems wrong, stmt number is not good enough for store names.

lib/transforms/type_inference.ml
```diff
@@ -0,0 +1,1598 @@
...
+    | Function _ -> (recursives, Top)
+
+  let rec type_to_inferred (typ : Types.t) : t =
+    match typ with
+    | Top -> Top
+    | Nothing -> Bottom
+    | Variable a -> TypeVar (VarId.make_id a)
+    | Pointer { lower; upper } ->
+        Pointer (type_to_inferred lower, type_to_inferred upper)
+    | Record (name, fields) -> failwith "TODO"
```
JTrenerry: finish this

JTrenerry:
extract has errors where the coalescing is wrong, i forget if i fixed this

JTrenerry:
loops almost type checks after inference

JTrenerry:
cntlm has inf loop somewhere

JTrenerry:
Apparently it is slow but I think it is trying its best :)

PR #60, opened 2026-02-23
Tree-sitter grammar and basic highlighting

katrinafyi:
You can see the docs to read all about it! https://github.com/agle/bincaml/blob/tree-sitter/docs/tooling.mld

image

### DIFF
--- a/docs/tooling.mld
+++ b/docs/tooling.mld
@@ -161,53 +161,60 @@ and the older
 {{: https://github.com/nvim-treesitter/nvim-treesitter/blob/master/README.md#adding-parsers }
 "master" version}.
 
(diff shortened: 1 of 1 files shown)


katrinafyi:
Real quick, it turns out that neovim with old nvim-treesitter also needs manual installing of highlights. 

Find the path with :echo nvim_list_runtime_paths() and symlink the file according to
> Nvim looks for queries as *.scm files in a queries directory under
runtimepath, where each file contains queries for a specific language and
purpose, e.g., queries/lua/highlights.scm for highlighting Lua files.

JTrenerry:
I think it is dependent on that kind of stuff, Helix doesn't set it automatically with due to paths being different for different systems.

agle (approved):

PR #62, opened 2026-02-24
Add semicolons on multiple require summaries

b-paul:
When dumping il on a procedure with multiple requires/ensures/rely/guarantees clauses, a semicolon is only added to the last clause instead of all lines, and this pr fixes that

e.g. it was emitting
```
  requires boolnot(boolnot(eq($_PC:bv64, 0x400d40:bv64)))
  requires eq($_PC:bv64, 0x400d40:bv64);
```


JTrenerry (approved):
LGTM

PR #63, opened 2026-02-25
add support for basil ir memassign stmt

### LINKED ISSUE #59: Parse error when global is written
Hi there, I tried running bincaml (0e2e8fc) on the attached IL file and I get:

```
bincaml: Error in (load-il write-global/simple.-O0-output.il): Parse error:  write-global/simple.-O0-output.il:11
         11 |     $Global_4325420_4325424: bv32 mem:= 0x2a:bv32 { .label = "4196176_0" };
                                                ^^^
          at Dune__exe__Script.of_cmd.(fun) bin/script.ml:54
```

This is the original C code:

```c
int x;

int main() {
	x = 42;
}
```

Thanks!

[write-global.zip](https://github.com/user-attachments/files/25477410/write-global.zip)

agle:
fix #59 

Add support for scalar memory access instructions.  We considered the following options: 

1. ~~add an observable flag to the variable declaration and use regular assignments; possibly replace load/store statements with expressions and assigns --- need to consider observability at expr level~~ This requires convoluted reasoning about obsservabilty at the expression level that we would rather avoid. 
2. Make the load/store instructions more general so they can express scalar assignments somehow
3. ~~Add another kind of load/store instruction specifically for scalar operations.~~ we have somewhat done this, but it is in the scope of (2); adding scalar/multi variants to the existing instructions. 

Adds the following statements:

```
Stmt_ScalarStore . Stmt ::= LVar ":=" "store" Expr;
Stmt_ScalarLoad . Stmt ::= LVar ":=" "load" Var ;
```

Modifies load and store statements to be optionally scalar:

```ocaml
  | Instr_Load of {
      lhs : 'lvar;  (** load destination variable *)
      rhs : 'var;  (** variable loading from *)
      addr : 'expr access;  (** load index *)
    }  (** a load from memory [rhs] at index [addr] *)
  | Instr_Store of {
      lhs : 'lvar;  (** store destination variable *)
      rhs : 'var;  (** store source variable *)
      value : 'expr;  (** value to store value (may be combined with [rhs]) *)
      addr : 'expr access;  (** store address *)
    }

type 'e access = Scalar | Addr of { addr : 'e; size : int; endian : endian }

```

Therefore

`Instr_Load {lhs; rhs; addr=Scalar}` has the same semantics as `lhs := rhs` but with observable side-effect of reading rhs. 

Similarly `Instr_Store {lhs; rhs; value}` means `lhs := value`, with side effect of reading lhs and rhs and modifying lhs.

### DIFF
--- a/lib/lang/program.ml
+++ b/lib/lang/program.ml
@@ -45,7 +45,7 @@ let pretty_declaration d =
         StringMap.find_opt "classification" attrib
         |> Option.to_list
         |> List.flat_map (function
-          | `Expr e -> [ text "classification " ^ Expr.BasilExpr.pretty e ]
+          | `Expr e -> [ text " classification " ^ Expr.BasilExpr.pretty e ]
           | _ -> [])
         |> append_l
       in
--- a/test/cram/roundtrip.t
+++ b/test/cram/roundtrip.t
@@ -1,10 +1,5 @@
 
   $ bincaml script roundtrip.sexp
-  bincaml: Error in (load-il beforemem.il): Parse error:  beforemem.il:1
-           1 | var observable $Global_4325420_4325424:bv32classification true;
-                                                      ^^^^^^^^^^^^^^^^^^
-            at Dune__exe__Script.of_cmd.(fun) bin/script.ml:64
-  [123]
 
 The serialise -> parse serialise loop should be idempotent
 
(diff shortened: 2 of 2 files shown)


agle:
Analyses will need updating to be precise for the scalar load/store.

PR #69, opened 2026-02-26
tree-sitter: fix proc colouring, add word, more tokens for fallback

katrinafyi:
in IL files with syntax errors (such as those using new IL features where the grammar hasn't been updated yet), tree-sitter will fall back to a token based  highlighting. it's important that we set the "word" token and we have highlighting rules for atomic tokens so that it still looks decent in these cases - otherwise, everything after the syntax error would be white.

todo: generate tree-sitter grammar.js in dune


JTrenerry:
none, and cant seem to find an option to make it wait longer

PR #70, opened 2026-02-26
Requires clause generation with WP dual domain

b-paul:
- Implements the wp dual domain to transfer assertions into requires clauses
- Add both an intraproc and interproc transform for adding requires clauses on all procedures
- Interproc solver performs fixedpoints on sccs of the callgraph (uses smt queries to compute leqs)

Currently lots of precision is lost on memory loads going to top.

### DIFF
--- a/lib/analysis/wp_dual.ml
+++ b/lib/analysis/wp_dual.ml
@@ -66,7 +66,7 @@ module Domain (S : RequiresAnnotation) = struct
                     (ApplyIntrin
                        {
                          attrib;
-                         op = `OR;
+                         op = `AND;
                          args =
                            vars
                            |> List.map


lib/transforms/function_summaries.ml
```diff
@@ -0,0 +1,204 @@
...
+
+    let is_maximal _ = false
+  end in
+  let module FixSummaries = Fix.Fix.ForHashedType (ID) (Domain) in
+  let eqs (pid : ID.t) (vals : FixSummaries.valuation) =
+    if ID.Set.mem pid component then
+      let annotations =
+        (module struct
+          let requires id =
+            List.append (ID.Map.find id res).requires (vals id).requires
```
agle: Use Containers_data.MultiMap

lib/transforms/function_summaries.ml
```diff
@@ -0,0 +1,204 @@
...
+      SMTLib2.assert_bexpr q SMTLib2.empty
+      |> snd
+      |> SMTLib2.to_sexp ~set_logic:false
+    in
+    let open Bincaml_util.Smt in
+    Solver.push solver;
+    s |> Iter.iter (fun c -> Solver.add_command solver c);
+    let res = Solver.check solver in
+    Solver.pop solver;
+    match res with Unsat -> true | Sat -> false | Unknown -> assert false
```
agle: Probably return false on unknown - or does that lead to divergence?

This probably needs a recovery strategy in case of timeout. Fix.lfp assumes finite height lattice though and doesn't support widening, so maybe truncate and make it clear this pass is incomplete with respect to widening the contract to include the behaviour of the procedure.

lib/transforms/function_summaries.ml
```diff
@@ -0,0 +1,204 @@
...
+
+let wp_dual_requires (module S : FunctionSummaryAnnotation)
+    (proc : Program.proc) =
+  let module Domain = Wp_dual.Domain (S) in
+  let module Analysis = Intra_analysis.Backwards (Domain) in
+  let result =
+    Analysis.analyse
+      ~init:(fun _ -> Expr.BasilExpr.boolconst false)
+      ~widening_set:Graph.ChaoticIteration.FromWto ~widening_delay:5 proc
+  in
```
agle: I should refactor analysis to pass a context value through of an arbitrary type declared in domain rather than requiring it be functorised.

lib/analysis/wp_dual.ml
```diff
@@ -0,0 +1,170 @@
...
+        | ApplyIntrin { op = `OR; args }
+          when List.mem ~eq:BasilExpr.equal e_true args ->
+            replace [%here] e_true
+        | ApplyIntrin { op = `AND; args }
+          when List.mem ~eq:BasilExpr.equal e_false args ->
+            replace [%here] e_false
+        | BinaryExpr { op = `EQ; arg1; arg2 } when BasilExpr.equal arg1 arg2 ->
+            replace [%here] e_true
+        | UnaryExpr { op = `BoolNOT; arg } when BasilExpr.equal arg e_true ->
+            replace [%here] e_false
```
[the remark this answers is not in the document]
b-paul: The idea for this simplifier was to have it specialised for the domain to allow it to keep predicates in a specific shape (later I might want to even not use the rewriter at all, instead working over top level expression terms only under the invariant that lower nodes are simplified).

agle (approved):

PR #71, opened 2026-02-27
Update tree-sitter after memassign and build in dune

katrinafyi:
todo

- [x] fix some typos around spec tokens
- [x] add spec tokens and classification tokens to highlight

# Big changes

Grammar is changed slightly:
- A block's phi nodes are now a separate parenthesised list _before_ the statement list. For example:
  ```c
  block %main_19 (
    var CF_2:bv1 := phi(%main_3 -> CF_6:bv1, %main_21 -> CF_1:bv1),
  ) [
    var R0_7:bv64 := 0x0:bv64;
    goto (%main_17);
  ]; 
   ```
  This is also in line with: https://github.com/UQ-PAC/BASIL/pull/591
- Function calls are done using a parenthesised comma-separated list of params, and whitespace-only function application (a la Ocaml/Haskell) is removed.
- Expr has learned precedence and associativity. In particular, quantified statements are right associative and have lowest precedence. This is followed by function application. The remaining "normal" expressions have highest precedence: binop, literals, etc.
  - The `store` statement has two expressions next to each other for address and value. Function application and quantified statements *cannot* appear unparenthesised as the address expression (for ambiguity reasons) (this is probably rare anyway).
  - BoolBinOp will now always parse as an AssocExpr rather than a BinExpr, even if it has exactly two operands.
- Changes the block printing to use 2 spaces instead of 3 (no difference in the grammar, but a lot of the cram tests have to be regenerated)

I strongly suspect that the printer will need to be changed to add parentheses where needed. So far, it has gotten away with assuming everything is unambiguous all the time, but no longer. This can be a later PR though. 

------
Building the tree-sitter needs the `bnfc-treesitter` package from pac-nix. This provides an updated `bnfc` program. Generating tree-sitter is only done if the environment variable `BINCAML_BNFC_TREESITTER=true` is set.
```
BINCAML_BNFC_TREESITTER=true dune runtest tree-sitter 
```

### DIFF
--- a/lib/fe/AbsBasilIR.ml
+++ b/lib/fe/AbsBasilIR.ml
@@ -117,7 +117,7 @@ and namedCallReturn =
 
 and lVars =
    LVars_Empty
- | LVars_LocalList of openParen * localVarParen list * closeParen
+ | LVars_LocalList of openParen * localVar list * closeParen
  | LVars_List of openParen * lVar list * closeParen
  | NamedLVars_List of openParen * namedCallReturn list * closeParen
 
@@ -135,7 +135,7 @@ and jump =
  | Jump_ProcReturn
 
 and lVar =
-   LVar_Local of localVarParen
+   LVar_Local of localVar
  | LVar_Global of globalVar
 
 and stmtWithAttrib =
--- a/lib/fe/BasilIR.cf
+++ b/lib/fe/BasilIR.cf
@@ -123,7 +123,7 @@ rules NamedCallReturn ::= LVar "=" LocalIdent;
 separator NamedCallReturn "," ;
 
 LVars_Empty . LVars ::= ;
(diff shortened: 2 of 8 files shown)


JTrenerry:
Is there anything sensible we can do to generate the tree-sitter source files when it has been updated?

agle:
I think no semicolon separating the contract list is good,


Does this parse with the change to type?


```
let $memory_load32_le : (bv64 -> bv8) -> bv64 -> bv32 = fun (#memory: bv64 -> bv8), (#index: bv64) ::
  (bvconcat(load_le(8, #memory, bvadd(#index, 3:bv64)),
    bvconcat((load_le(8, #memory, bvadd(#index, 2:bv64))),
    bvconcat((load_le(8, #memory, bvadd(#index, 1:bv64))),
    load_le(8, #memory, #index)))));
```

katrinafyi:
Yeah of course

PR #77, opened 2026-03-04
ADT type

agle:
Add ADT type and uninterp sort matching SMTlib


agle:
CVC5 supports polymorphic and mutually recursive type definitions; I haven't really thought about how to do that here. We can have definitions that are recursive but the references to the recursive type inside the recursion are just encoded as an opaque sort; textually that should translate to SMT correctly and I can make type equality nominal so that the type checker works out.  I have the syntax to write mutually recursive type definitions ocaml style but they don't get translated to a group of mutually recursive definitions in the IR/smt; should be easy.

I don't see a use for polymorphic type declarations unless we start writing polymorphic specifications, which is only very useful if we go as far as adding polymorphism to functions and procedures, which would be much more of a pain.

PR #79, opened 2026-03-08
Nix: nixpkgs-based flake, instead of dune-/opam-nix

katrinafyi:
Closes https://github.com/agle/bincaml/issues/54

This uses the nixpkgs buildDunePackage (https://ryantm.github.io/nixpkgs/languages-frameworks/ocaml/), instead of opam-nix or dune-nix. This is a lot simpler and it gives good separate caching of dependencies and still lets you use ocaml compiler options. However, it's fully manual and you'll always need to add the dependencies to bincaml.nix manually.

I've tried it out in `nix develop` and it seems to work well. `dune build` can build the project and successfully finds the dependencies installed by nix. I've also set up some environment variables which are needed to make odig work inside the nix shell.

Let me know if this works for you and whether it improves on the situation at the moment.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -64,8 +64,9 @@
           };
 
           devShells = {
-            default = selfOcamlPackages.callPackage ./nix/shell.nix { };
+            default = self.devShells.fp;
             fp = fpOcamlPackages.callPackage ./nix/shell.nix { };
+            no-fp = selfOcamlPackages.callPackage ./nix/shell.nix { };
           };
         }
       ;
--- a/nix/README.md
+++ b/nix/README.md
@@ -13,13 +13,16 @@ added to bincaml, this is the place they must be added.
 There are also useful dev shells for working on the project. These
 provide the build-dependencies for bincaml, as well as useful tools
 like the ocaml-lsp.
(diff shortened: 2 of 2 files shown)


agle:
Why would you ever need the build without frame pointers? There should not be any difference in build time between the two? Does it enable flambda? We want the switch to have flambda enabled but not build with flambda by default.

katrinafyi:
stopped... at the last second :sob:  build wo fp can re-use the nixpkgs cache so it just downloads all the prebuilt ocaml packages and doesn't need to build them.

agle:
That should only be a problem for the first build right? Or does nix force you to rebuild the whole switch every time?

agle:
Dev build does not enable flambda (`-O0` vs `-O3`)- we don't need to be developing on a release build. I.e. I just mean install the compiler that supports flambda without using it.

katrinafyi:
> That should only be a problem for the first build right? Or does nix force you to rebuild the whole switch every time?

Yeah, it is only a problem for the first build. I personally want a build without fp most of the time because i'm not usually in the performance space.

PR #88, opened 2026-03-12
Interprocedural dead store elimination

b-paul:
Uses IDE live results to do dead store elimation. Dead parameters are first removed, then dead stores based on which parameters were removed. There is a test case included that shows this interprocedural transform is better in at least one scenario.

The intraproc transform isn't currently updated to handle phi nodes, so we perform the transform before ssa. The number of lines in the ssa -> interproc transform is around 13k lines less than that of intraproc -> ssa for cntlm! (line count is somewhat inflated because of the pretty printer collapsing lines, but it shows it's doing stuff still)

A further modification of the ide solver was made to generate useful phase 2 results for this transform, where the analysis can specify default values at the "beginning" of each procedure! Here we specify that each out var is live.

Also note the ide-live transform was removed

### DIFF
--- a/lib/analysis/idessi.ml
+++ b/lib/analysis/idessi.ml
@@ -80,28 +80,32 @@ module IDESSI (D : IDESSIDomain) = struct
     let compare = Ord.triple DL.compare ID.compare DL.compare
   end)
 
-  let propagate worklist summaries get_summary update_summary updates =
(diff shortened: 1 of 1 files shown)


b-paul:
On further reflection, not eliminating input variables that are used in requires clauses isn't so easy as liveness is propagated backwards... Perhaps another analysis is needed

PR #91, opened 2026-03-16
Boogie Backend

mira-alford:
Initial support for a boogie backend.

Seems to pass boogies parsing + typechecking and verify (or fail to verify) on all of the examples, plus some additional tests for features like quantifiers that are not present.

Use the `(dump-boogie "out.bpl")` expression to generate boogie output.

### DIFF
--- a/lib/backends/dune
+++ b/lib/backends/dune
@@ -3,5 +3,4 @@
  (name backends)
  (flags -w -27)
  (modules boogie)
- (libraries containers containers-data containers.pp lang transforms)
- (inline_tests))
+ (libraries containers containers-data containers.pp lang transforms))


lib/transforms/type_check.ml
mira-alford: @JTrenerry lmk if this type checking is fine

lib/transforms/cf_tx.ml
```diff
@@ -4,27 +4,47 @@ open Bincaml_util.Common
...
+      ensures = List.map (rewriter ?visit) s.ensures;
+      rely = List.map (rewriter ?visit) s.rely;
+      guarantee = List.map (rewriter ?visit) s.guarantee;
+    }
+  in
+  Procedure.set_specification p s
+
+let simplify_prog_spec_exprs rewriter ?visit (p : Program.t) =
+  let procs =
+    ID.Map.map (fun proc -> simplify_proc_spec_exprs rewriter ?visit proc) p.procs
```
JTrenerry: Can have it as

`(simplify_proc_spec_exprs rewriter ?visit)` to have less readable but cooler code!

PR #103, opened 2026-03-25
convert smt to expr

agle:
- supports predicates and bvexprs that can be generated by bincaml
- will not support all bvops defined in smtlib, we could extend it but it will need some translation into the IR (e.g. we don't define greater-than operators,  so convert to not-lessthan).
- expands expr gen property test to generate predicates
- removes associative binary operators from Ops.AllOps.binary so they are only present in Ops.AllOps.intrin -- make their representation canonical


lib/analysis/defuse_bool.ml
```diff
@@ -116,12 +116,15 @@ module IsZeroValueAbstraction = struct
...
   let eval_intrin (op : Lang.Ops.AllOps.intrin) (args : t list) =
     match op with
-    | `BVADD -> List.fold_left (eval_binop `BVADD) Bot args
-    | `BVOR -> List.fold_left (eval_binop `BVOR) Bot args
-    | `BVXOR -> List.fold_left (eval_binop `BVXOR) Bot args
-    | `BVAND -> List.fold_left (eval_binop `BVAND) Bot args
+    | `BVADD -> List.fold_left (eval_binary `BVADD) Bot args
+    | `BVOR -> List.fold_left (eval_binary `BVOR) Bot args
+    | `BVXOR -> List.fold_left (eval_binary `BVXOR) Bot args
+    | `BVMUL -> List.fold_left (eval_binary `BVXOR) Bot args
```
b-paul: eval_binary `BVMUL?

b-paul (approved):
Looks good (except for the one really small thing I noted), but I'd expect at least one thing to break if main is merged into this from binops being removed and turned into intrinsics only (at least one of my prs used them).

If you're busy I'm happy to handle the conflicts and whatnot, up to you

PR #111, opened 2026-04-07
Interprocedural copy propagation

b-paul:
Copy propagation. It's very fast and uses a funky algorithm (described in a comment of the code).

Importantly we are able to copy through phi nodes, in the sense that if x = phi(y, z) and y and z are both copied from w, then x is.

Interprocedurality comes into play in two ways (second implies the first). The first way is if an output variable is a copy of one input, then we can copy through a call. The second way is if an output variable is a phi of some input variables, and each of those input variables in a call are the same, in which case the copy should still propagate).

The transform cuts like 40k loc on cntlm if you apply dse after copy propagating, compared to just applying dse (all on constant folded ssa). It also runs in ~280ms on my slow laptop!


mira-alford (approved):
Looks good to me 🚀

PR #113, opened 2026-04-08
Fixes parser tests

JTrenerry:
Tests broke in #102 and for some reason the cram does not like looking at new files anymore.


katrinafyi:
that's devastating

agle:
- `record = Record of {a : bv64; b: bv32; c:bv64}` should not be a valid type annotation the parser should reject it; we should have a separate declaration type expression and regular type expression
- `record` is the name of the type so is the intended type to attach to the expression, if we want to inline all type definitions to their uses I think we should do so in a subsequent elaboration pass

JTrenerry:
@katrinafyi this removes a set of brackets around records like we talked about, you still want?

JTrenerry:
@agle this also changes the to_string of sorts to be the name and not the entire thing, not sure if this works though, I assume it actually needs something supporting it?

lib/lang/program.ml
```diff
@@ -125,6 +125,13 @@ let prog_pretty (p : t) =
   let open Containers_pp.Infix in
   let globs =
     StringMap.bindings p.globals
+    |> List.sort (fun (_, decl) (_, decl2) ->
+        (* NOTE: Recursive types might require more logic here *)
+        match (decl, decl2) with
+        | Type _, Type _ | Variable _, Variable _ | Function _, Function _ -> 0
+        | Type _, _ -> -1
+        | _, Type _ -> 1
+        | _ -> 0)
```
katrinafyi: generally, you should try to avoid manually writing compare functions because it's error prone.

instead, you can use https://c-cube.github.io/ocaml-containers/containers/CCOrd/index.html to compose them. in this case, you could use ccord map to map each decl into 0 1 or 2 based on its kind, then use ccord.int to use the derived integer order.

katrinafyi (commented):
> @katrinafyi this removes a set of brackets around records like we talked about, you still want?

yeah that's fine

lib/lang/program.ml
JTrenerry: I am not sure if this is needed anymore, ill test it and see, but pretty sure this was a fix that never worked lol

test/cram/ptrrec1.il
```diff
@@ -1,25 +1,23 @@
...
 proc @main_4196164(R0_in:bv64, R10_in:bv64, R11_in:bv64, R12_in:bv64, R13_in:bv64,
    R14_in:bv64, R15_in:bv64, R16_in:bv64, R17_in:bv64, R18_in:bv64, R1_in:bv64,
    R29_in:bv64, R2_in:bv64, R30_in:bv64, R31_in:bv64, R3_in:bv64, R4_in:bv64,
    R5_in:bv64, R6_in:bv64, R7_in:bv64, R8_in:bv64, R9_in:bv64, _PC_in:bv64)
    -> (R0_out:bv64, R1_out:bv64) { .address = 4196164; .name = "main";
     .returnBlock = "main_return" }
-  modifies $rec:{("0": (bv32, 0))}
-  captures $rec:{("0": (bv32, 0))}
+  modifies $rec:{"field0": (bv32, 0), "field1": (bv64, 32)}, $rec:{"0": (bv32, 0)}
+  captures $rec:{"field0": (bv32, 0), "field1": (bv64, 32)}, $rec:{"0": (bv32, 0)}
```
agle: This looks wrong, we have two non-equal global variables of the same name but different types because the type annotation on the value in the body is different; should probably be a type error.

JTrenerry: yeah, forgot to fix one of the types in the actual code, so that was being added in.

agle (approved):

PR #115, opened 2026-04-09
SVA

JTrenerry:
Remove SVA from #58 and move it here.

Very limited usage and testing for it isn't amazing at the moment, but from limited practice with malloc results, it does function.

TODO: have at least one expect test


b-paul (commented):
probably good other than these changes

b-paul (approved):

JTrenerry:
Fix for this is in #58, doing soon but I BROKE EVERYTHING :sob:

PR #116, opened 2026-04-09
Bincaml-based LSP

katrinafyi:
This is a basic LSP which has:
- go to definition of block/global/parameter variables, blocks, procedures, ADT cases, and declared type names.
- hover information of the above terms,
- autocompletion of the above terms,
- symbol tree view,
- procedure DOT graph generation,
- syntax error highlighting from lexer / parser,
- loading error highlighting (e.g., due to undefined variable errors),
- debug highlight mode where every token is highlighted,
- an attempt at a cached on-demand dependency system for internal LSP state.

You will need to configure the LSP in your editor for the appropriate file type. In Astronvim, it's something like
```lua
    config = {
      basillsp = {
        cmd={"/home/rina/progs/obasil/_build/default/lsp/bin/main.exe"}, filetypes={"basilir"}, root_dir = require("lspconfig.util").root_pattern("."),},

    },
```
inside the `~/.config/nvim/lua/plugins/astrolsp.lua` file.

The DOT graph generation and highlighting mode is accessed using LSP code actions. In Astronvim, it's `<Space>la`.

### Hover information

  Screenshot_20260409_163445
  Screenshot_20260409_163620

### Symbol outline
image

### Error and token highlighting

image
image
image

### Code actions

image


lsp/bin/main.ml
```diff
@@ -0,0 +1,364 @@
...
+              | Some graph ->
+                  let file_name = Filename.temp_file "bincaml_graph." ".dot" in
+                  Out_channel.with_open_text file_name (fun oc ->
+                      let fmt = Format.formatter_of_out_channel oc in
+                      Lang.Viscfg.Dot.fprint_graph fmt graph;
+                      Format.pp_print_flush fmt ());
+                  let open Lwt.Syntax in
+                  let* () =
+                    run_command ~notify_back
+                      (Filename.quote_command "dot"
```
JTrenerry, line 160: so lsp requires dot?

katrinafyi: Yeah, but only for the generate graph code action.

JTrenerry: ok, xdg-open is pre cool, did not know about it

katrinafyi: yeah, won't work on macos tho ;-;

JTrenerry: w

agle: Macos should have "open" https://github.com/b0-system/b0/blob/63c4db467ab3ad34c8e8744269915947436c5601/src/std/b0_web_browser.ml#L48

JTrenerry:
flake doesn't have some of the packages needed, although my nix is dying right now so unsure if skill issue

agle:
Go to definition of globals goes to declaration rather than last definition

lsp/bin/index.mld
```diff
@@ -0,0 +1,32 @@
+
+{0 Basil IR Language server }
+
+
+This is a simple language server for the {{: https://github.com/UQ-PAC/BASIL} Basil}
```
agle: This isn't really the basil intermediate language anymore, are we still a superset of that?

lsp/bin/index.mld
```diff
@@ -0,0 +1,32 @@
...
+intermediate language.
+
+It provides the following basic functionality:
+
+- syntax checking
+- goto-definition for block and procedure labels
+- symbol list
+- semantic highlighting
```
agle: generate graph, autocomplete

lsp/bin/main.ml
```diff
@@ -0,0 +1,364 @@
...
+
+   Here we expect a few things:
+   - a type to represent a state/environment that results from processing an
+     input file
+   - a function processing an input file (given the file contents as a string),
+     which return a state/environment
+   - a function to extract a list of diagnostics from a state/environment.
+     Diagnostics includes all the warnings, errors and messages that the processing
+     of a document are expected to be able to return.
+*)
```
agle: Delete placeholder comment

lsp/bin/main.ml
```diff
@@ -0,0 +1,364 @@
...
+                       not (Bincaml_lsp.Raw_tokens.lsprange_contains range pos))
+                   |> Option.value ~default:true)
+          in
+          Lwt.return (Some (`List completions))
+
+    method! config_symbol = Some (`Bool true)
+
+    method! on_req_symbol ~notify_back ~id ~uri ~workDoneToken
+        ~partialResultToken () =
+      Printexc.record_backtrace true;
```
agle: Is this intentional?

lsp/raw_tokens.ml
```diff
@@ -0,0 +1,333 @@
...
+  | KW_boolnot
+  | KW_boolimplies
+  | KW_booland
+  | KW_block
+  | KW_be
+  | KW_axiom
+  | KW_assume
+  | KW_assert
+  | KW_and
+[@@deriving show { with_path = false }, eq]
```
agle, line 2: ppx_import ?

lsp/lsp_symbols.ml
```diff
@@ -0,0 +1,276 @@
...
+         (decl.range.start, { decl with range } :: rest))
+       ~init:(end_, [])
+  |> snd
+
+(* HACK: we have different kinds of toplevel decls, but we have to fit them into
+   the lsp types. in future, maybe we make our own symbol type... *)
+
+let proc_lspsymbol_at_pos ~lspsymbols lsppos =
+  lspsymbols
+  |> List.find_opt (fun (sym : symbol) ->
```
agle, line 247: Performance improvement available by using Map.split

lsp/lsp_symbols.ml
```diff
@@ -0,0 +1,276 @@
...
+      match sym.kind with
+      | Enum | Interface -> false
+      | _ -> Raw_tokens.lsprange_contains sym.range lsppos)
+
+let lspsymbols_with_kind ~lspsymbols ~lsppos (kind : ident_kind) : symbol Iter.t
+    =
+  let symbol = lazy (proc_lspsymbol_at_pos ~lspsymbols lsppos) in
```
agle: This doesn't seem to be reused so could just go inside locals ()

lsp/lsp_state.ml
```diff
@@ -0,0 +1,239 @@
...
+                    ( `MarkupContent
+                        {
+                          Linol_lsp.Types.MarkupContent.kind = Markdown;
+                          value = message;
+                        },
+                      None )
+              in
+              let range = Option.value ~default:default_range range in
+              [ Lsp.Types.Diagnostic.create ~message ~range ~severity:Error () ]
+        in
```
agle, line 173: Mentioned previously but including here:

- do we want to use algebraic effects to pass errors through

```ocaml

effect Warning of string * Lexing.pos * filename (*display and continue *)
effect Alarm of string * Lexing.pos  * filename (* display and continue *)
exception Error (*give up*)

```

- is there a way to do this with a Logs reporter that includes a lexing position tag and displays the output as lines

lsp/lsp_symbols.ml
```diff
@@ -0,0 +1,276 @@
...
+    (match block with
+      | Block_NoPhi (bid, _, _, stmts, _, _) ->
+          Iter.of_list stmts
+          |> Iter.flat_map (function StmtWithAttrib1 (stmt, _) -> of_stmt stmt)
+      | Block_Phi (bid, _, _, phis, _, _, stmts, _, _) ->
+          Iter.of_list stmts
+          |> Iter.flat_map (function StmtWithAttrib1 (stmt, _) -> of_stmt stmt))
+    |> Iter.map (fun ((range, name), context) ->
+        ((range, name), Field, context @ ("within block" :: bcontext)))
+  in
```
agle, line 136: In general I think i would prefer a solution that embeds the neccessary information in bincaml ir with the minimal amount of bookeeping e.g. the token under cursor map. We still have work to do to properly support everything and its inherently lossy so its something to work towards in the future. `loadir.ml` is a mess and needs some substantial work to make it cleaner and separate its various concerns (esp. type inference, processing concrete ast,  and ir building)

agle (commented):
There's nothing really objectionable here, its very tightly coupled to the concrete ast from bincaml which might produce some maintenence churn but we'll see. needs some docs

agle (changes requested):
(docs changes required from previous review)

katrinafyi:
> Go to definition of globals goes to declaration rather than last definition

With this, did you mean like a reassignment to `$mem`? Is there an example?

agle:
Any register assignment where the register is a global will go to the declaration in global scope rather than the previous assignment. Not sure if LSP ui has support for multiple and it requires a control flow analysis so maybe ignore.

katrinafyi:
Yeah that's expected atm but it would be really useful. I can definitely look into it after this PR.

lsp/lsp_symbols.ml
katrinafyi: Yeah it's true but this only iterates through toplevel declarations so I'm gonna leave it for now. There's other O(n) work happening anyway.

lsp/lsp_state.ml
katrinafyi: It would be cool. I think Logs integration is somewhat easier if we don't need to dynamically choose between continuing or aborting. We can do this later.

lsp/lsp_symbols.ml
katrinafyi: Yeah :)

katrinafyi (commented):
comments should be fixed now.

katrinafyi:
Has anyone seen this before?
```
Promoting                              
  _build/default/bincaml-lsp/lib/ocaml/5.4.1/site-lib/bincaml_lsp/.formatted/bincaml_lsp.ml
  to bincaml-lsp/lib/ocaml/5.4.1/site-lib/bincaml_lsp/bincaml_lsp.ml.
Error: failed to promote               
bincaml-lsp/lib/ocaml/5.4.1/site-lib/bincaml_lsp/bincaml_lsp.ml
Permission denied
```

ohh this happens when you have a `nix build .` output in the current folder.

dune-project
```diff
@@ -50,11 +50,27 @@
...
   (ppx_expect
-   (>= "0.17.2.2"))
+   (>= 0.17.2.2))
   (alcotest :with-test)
-  (qcheck-core :with-test)
-  (qcheck-alcotest :with-test)
-  (qcheck-stm :with-test))
+  qcheck-core
+  qcheck-alcotest
+  qcheck-stm)
```
agle: Is this intentional?

katrinafyi: There was an `executable` which depended on them so opam-dune-line thought it was required outside of test. Changed in 38e96ed.

katrinafyi: Fixed better in 5444667. You can read about it in the commit message.

agle (approved):

JTrenerry:
YAY

PR #118, opened 2026-04-10
ci: cache on github actions and fix `dune @fmt` rule

katrinafyi:
approach is following the idea of https://discuss.ocaml.org/t/cache-with-gh-action-ocaml-setup-ocaml-v2/11650/7. we also use  [composite actions](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action) to factor it out from each workflow file

cached build times are down to 1-2 minutes, down from 6+ minutes.

### DIFF
--- a/.github/actions/bincaml-setup/action.yml
+++ b/.github/actions/bincaml-setup/action.yml
@@ -11,10 +11,14 @@ runs:
       with:
         ocaml-compiler: ${{ inputs.ocaml-compiler }}
 
+    - id: month
+      run: python3 -c 'from datetime import *; print(datetime.now().strftime("month=%m"))' >> "$GITHUB_OUTPUT"
+      shell: bash
+
     - uses: actions/cache@v5
       with:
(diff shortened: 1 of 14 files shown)


katrinafyi:
i'm losing it. in ci, we have
```
[NOTE] Package ocamlformat is already installed (current version is 0.29.0).
```
i have the same one on my computer.

in ci, if we do `opam exec -- dune build @fmt`, it shows errors. on my computer, if we do that command, it shows no errors.

JTrenerry:
Maybe it is time for

>Warning: No ocamlformat version found in .ocamlformat file. It's recommended to specify the version in your .ocamlformat file for better consistency.

agle:
I tried docker and its only marginally faster https://github.com/agle/bincaml/actions/runs/24381285654/job/71205114682

katrinafyi:
Abandoning after the container-based solution in https://github.com/agle/bincaml/pull/122

PR #119, opened 2026-04-13
Logs

JTrenerry:
# What it do:
Print logs based on levels
```
; (log-level "quiet")
; (log-level "app") (* probably useless for us *)
; (log-level "error")
; (log-level "warning")
; (log-level "info")
; (log-level "debug")
```

Note: Will need #113 merged first since I made it off of that branch :sob:


bin/main.ml
```diff
@@ -99,6 +100,8 @@ let cmd =
 let main () =
   Trace_core.set_process_name "main";
   Trace_core.set_thread_name "t1";
+  Logs.set_level (Some Logs.Error);
+  Logs.set_reporter (Logger.reporter Format.std_formatter);
```
katrinafyi: probably err formatter instead right? in case stdout gets used for piping in the future.

bin/script.ml
```diff
@@ -66,6 +66,11 @@ let of_cmd st (e : Containers.Sexp.t) =
   Trace_core.with_span ~__FILE__ ~__LINE__ ("runcmd::" ^ cmd) (fun _ ->
       match cmd with
       | "skip" -> st
+      | "debug-level" ->
+          let level = List.hd (assert_atoms 1 args) in
+          let open Result in
+          Logs.set_level @@ Option.get_or ~default:None @@Result.to_opt @@ Logs.level_of_string level;
```
katrinafyi: Common.ReplError

lib/fe/BasilIR.cf
```diff
@@ -9,11 +9,11 @@ position token INTTYPE {"int"} ;
 position token BOOLTYPE {"bool"} ;
 position token POINTERTYPE {"ptr"} ;
 
-{- position token UserIdent '`' ((upper | letter | '_' | '#')(upper | letter | digit | ["_.$#"])*) '`'; -}
-position token BIdent '.'((upper | letter | '_')(upper | letter | digit | ["_.$#~"])*) ;
+{- position token UserIdent '`' ((upper | letter | '_' | '#')(upper | letter | digit | ["_$#"])*) '`'; -}
+position token BIdent '.'((upper | letter | '_')(upper | letter | digit | ["_$#~"])*) ;
```
katrinafyi, line 13: change is unrelated but NBD ig

bin/script.ml
```diff
@@ -66,6 +66,11 @@ let of_cmd st (e : Containers.Sexp.t) =
   Trace_core.with_span ~__FILE__ ~__LINE__ ("runcmd::" ^ cmd) (fun _ ->
       match cmd with
       | "skip" -> st
+      | "debug-level" ->
```
katrinafyi: "log-level" makes it a bit more generic, rather than having debug-level debug

katrinafyi (approved):

lib/fe/BasilIR.cf
JTrenerry: Yeah, I accidentally made this off of the #113 branch :sob:

bin/script.ml
JTrenerry: :+1:

bin/script.ml
JTrenerry: What does this do?

JTrenerry: nevermind, i got it

katrinafyi: yeah cool. you also may have to set all:true to make the change affect older log sources 

https://erratique.ch/software/logs/doc/Logs/index.html#val-set_level

nvm it's the default 🙂‍↕️

JTrenerry:
Logs seems like a bad library to use, will switch to https://github.com/codinuum/volt/tree/master as it seems to support what we need.

Edit: nvm not worth the effort to build

JTrenerry:
I think tag / src based logs is probably irrelevant and can be replace by just moving when you set the log level.

i.e. if you only want type-inference logs do
```
(load-il "examples/cntlm-output.il")
(run-transforms "cf-expressions")
(run-transforms "ssa")
(run-transforms "cf-expressions")
(run-transforms "simplify")
(dump-il "before.il")
(log-level "info")
(run-transforms "type-inference")
(log-level "quiet")
```
not
```
(load-il "examples/cntlm-output.il")
(log-level "info")
(run-transforms "cf-expressions")
(run-transforms "ssa")
(run-transforms "cf-expressions")
(run-transforms "simplify")
(dump-il "before.il")
(run-transforms "type-inference")
(log-level "quiet")
; (dump-il "after.il")

```

PR #122, opened 2026-04-14
ci: workflows based on container

agle:
These include bnfc, cvc5, boogie, and tree-sitter-cli so run all the tests, and introduce the requirement that treesitter is kept up-to-date.

Also only run on 5.3

- action `lint-doc` just ran dune build @doc, so we assume that is subsumed by `Build odoc` and remove it.

Only a subset of the packages are in the container so there is still a significant build time.


tree-sitter/queries/highlights.
```diff
@@ -0,0 +1 @@
+/home/alicia/.config/nvim/queries/basilir/highlights.scm
```
katrinafyi:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,66 @@
+name: build and test linux
```
katrinafyi: can this just be in test.yml, as a separate job?

tree-sitter/highlights.scm
```diff
@@ -0,0 +1 @@
+/home/alicia/.config/nvim/queries/basilir/highlights.scm
```
katrinafyi:
```suggestion
```

tree-sitter/dune
```diff
@@ -35,7 +35,7 @@
    (diff? grammar.bnfc.js tree-sitter-basilir/grammar.js))))
 
 (rule
- (alias runtest)
+ (alias treesit)
```
katrinafyi, line 40: What's the need for this?

agle: To make it a separate complaint in the CI rather than just amongst the rest of the test logs, and so it doesn't make the rest of the tests fail and I can exclude it from the list of CI checks required for merge.

katrinafyi: I see. I did like treesitter running automatically in runtest because it would always remind me locally, but yeah I can't think of a nice way to exclude this from CI checks.

agle:
> [nix users need to update bnfc!! ](https://github.com/agle/bincaml/pull/122/commits/00ba5361b78ba38df7ab345dbb72657b2259b437)

Is this just a matter of changing a url or sha in the flake?

agle:
Did you forget to commit `dune-project`?

katrinafyi:
> > [nix users need to update bnfc!! ](https://github.com/agle/bincaml/pull/122/commits/00ba5361b78ba38df7ab345dbb72657b2259b437)
> 
> Is this just a matter of changing a url or sha in the flake?

i've updated the flake in this pr, so nix users should pick it up next time they start a dev shell.

PR #123, opened 2026-04-14
log source filtering, and accept `-` as script file

### LINKED ISSUE #65: Add logging and debug configuration
We need some organised logging system, likely use https://github.com/dbuenzli/logs

This needs to be integrated with verbosity flags, as well as flags for setting debug levels. 

- logging should have the ability to switch on logs in specific contexts; e.g. for specific analyses to make it easy to filter
- debug levels should be configurable to enable different assertions and potentially write debug info to files etc

katrinafyi:
log format looks like this now, without timestamps
```
[INFO] (application) Starting irreducible-loops
[INFO] (analysis.irreducible_loops) found 15 loops, 1 irreducible
[INFO] (analysis.irreducible_loops) found 0 loops, 0 irreducible
```

timestamps would cause trouble with cram tests.

the log-level sexp command is changed to accept one or more source names after the log level. 
```
(log-level "info")
(log-level "error" "analysis.irreducible_loops")
```

Closes https://github.com/agle/bincaml/issues/65

i don't know why some other cram tests have changed...

### DIFF
--- a/bin/main.ml
+++ b/bin/main.ml
@@ -7,7 +7,7 @@ open Cmdliner.Term.Syntax
 let () = Printexc.record_backtrace true
 
 let fname =
-  let doc = "Input file name (filename.il)" in
+  let doc = "Input file name (e.g., filename.il or - for stdin)" in
   Arg.(required & pos 0 (some string) None & info [] ~docv:"FNAME" ~doc)
 
 let proc =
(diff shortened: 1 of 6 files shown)


agle (approved):

agle:
We've really gotta do something about the boogie output order swapping every pr

PR #127, opened 2026-04-16
refactor program module

agle:
- hide the type of program and introduce functions to access (this will make future refactors easier)
- make all declarations share a common id generator and map (this should move complexity out of loadir.ml)
   - also means we get a stable iteration order of declarations based on creation order from the id generator (though it is difficult/impossible to explicitly set the order)
- a bunch of other helper functions in Program
- reorganise code in ssa.ml for readability
- fix #104


bin/script.ml
```diff
@@ -158,14 +156,9 @@ let of_cmd st (e : Containers.Sexp.t) =
           in
           CCIO.with_out ofile (fun c ->
               let id =
```
b-paul: probably should be called p

lib/backends/boogie.ml
```diff
@@ -470,60 +424,93 @@ let pretty_procedure_spec (p : Lang.Program.proc) =
...
-  let glob_funs = append_nl glob_funs in
-  let procs =
-    p.procs |> IDMap.to_list
-    |> List.map (fun (_, p) -> pretty_procedure p)
-    |> append_l ~sep:(newline ^ newline)
-  in
-  append_l ~sep:(newline ^ newline) [ glob_vars; glob_funs; procs ]
+  let glob_vars = append_nl glob_vars_funs in
+  let rest = append_nl rest in
+  append_l ~sep:(newline ^ newline) [ glob_vars; rest ]
```
b-paul, line 511: is procs meant to be deleted here? it looks no but i might be missing something

lib/fe/BasilIR.cf
```diff
@@ -300,13 +299,13 @@ SortValRec . Expr ::= LocalIdent BeginRec [FieldAssign] EndRec;
...
 rules BVUnOp ::= "bvnot" | "bvneg" ;
-rules BVBinOp ::= "bvand"|"bvor"|"bvadd"|"bvmul"|"bvudiv"|"bvurem"|"bvshl"|"bvlshr"|"bvnand"|"bvnor"
-  |"bvxor"|"bvxnor"|"bvcomp"|"bvsub"|"bvsdiv"|"bvsrem"|"bvsmod"|"bvashr" ;
+rules BVBinOp ::= "bvudiv"|"bvurem"|"bvshl"|"bvlshr"|"bvnand"|"bvnor"
+  |"bvxnor"|"bvcomp"|"bvsub"|"bvsdiv"|"bvsrem"|"bvsmod"|"bvashr" ;
 rules BVLogicalBinOp ::= "bvule" |  "bvugt" |  "bvuge" | "bvult" | "bvslt" |  "bvsle" |  "bvsgt" |  "bvsge" ;
 
 rules IntBinOp ::= "intadd" | "intmul" | "intsub" | "intdiv" | "intmod"  ;
 rules IntLogicalBinOp ::=  "intlt" | "intle" | "intgt" | "intge" ;
-rules BoolBinOp ::=  "booland" | "boolor" | "boolimplies" ;
```
b-paul, line 309: do we want implication removed? i would have thought it would be nice for writing specs

b-paul (commented):
these changes look very nice yay

lib/backends/boogie.ml
agle: yeah its in the declarations list now

b-paul (approved):

PR #128, opened 2026-04-17
Interprocedural linear expression propagation

b-paul:
This transform propagates assignments that are linear functions (a * x + b) through their uses. For example, if `R31_2 := R31_1 - 0x2c0` then every use of `R31_2` would be replaced by the expression `R31_1 - 0x2c0` (plus transitivity).

For this transform to be a superset of copy propagation we also need to do a simultaneous copy only analysis, so that phi nodes are properly propagated into (since phis can only have variables in them, not linear expressions). Since it's a superset the old copyprop was removed.

The transform is interprocedural and super fast (700ms)!! I also added some stub stuff to make it propagate R31s on stubs properly !! (deleted abort case text) it handles exits too!!!!

Interestingly also the transform spends a lot of time doing wellformedness checks (the blank part)
image


lib/transforms/linear_copy.ml
```diff
@@ -0,0 +1,55 @@
...
+           ~f_rvar:copied_by)
+        block)
+    proc
+
+let transform (p : Program.t) =
+  let gs = Solver.solve p in
+
+  Trace_core.with_span ~__FILE__ ~__LINE__ "transform" @@ fun _ ->
+  Program.map_procedures
+    (fun pid proc -> transform_proc p (Hashtbl.find gs pid) proc)
```
mira-alford: Empty procedures like this:
```il
proc @shadow(x:bv64,y:bv64) -> () {}
[
	block %shadow_entry [
		goto (%shadow_return);
	];
	block  %shadow_return [
		return();
	];
];
```

Have no internal variables and cause the hashmap lookup to crash. Should probably just skip such procedures given there is nothing to copy/transform anyway?

mira-alford (approved):
Aside from that crash on completely empty procedures, LGTM

PR #129, opened 2026-04-17
dynamic single assignment transformation for lowering phi nodes

### LINKED ISSUE #135: Mismatched call/return argument/lhs counts throw CCList.Combine error
```
prog entry @f;

proc @f(x:bv64) -> (o:bv64)
[
    block %f_entry [
        (var y:bv64, var z:bv64) := call @f(x:bv64, x:bv64);
        return (y:bv64, z:bv64);
    ];
];
```
throws `Invalid_argument("CCList.combine")`. Fixing this is a little annoying since the code that calls List.combine doesn't have source code positions

katrinafyi:
incidental changes elsewhere:
- better error message when return statement and formal out params differ
- comments in procedure


lib/transforms/dsa.ml
```diff
@@ -0,0 +1,109 @@
...
+    DSA-style assignments performed in the new intermediate block. At the same
+    time, removes old direct edges from [src] to [tgt]. *)
+let replace_phis_with_dsa_blocks procedure (phis : dsa_block Iter.t) =
+  Iter.fold
+    (fun procedure { src; tgt; phi_assignments; assumes } ->
+      let phi_assign =
+        Stmt.Instr_Assign
+          (phi_assignments
+          |> List.map (fun (lhs, rhs) ->
+              (lhs, Expr.BasilExpr.E (RVar { attrib = None; id = rhs }))))
```
agle, line 57: Should use fix, the type should probably be hidden

lib/transforms/dsa.ml
```diff
@@ -0,0 +1,109 @@
...
+  |> Iter.filter_map (function
+    | Procedure.Vert.End src, Procedure.Edge.Jump, Procedure.Vert.Begin tgt -> (
+        match block_of_id tgt with
+        | { phis = []; stmts } -> None
+        | { phis; stmts } ->
+            let assumes, rest =
+              CCList.take_drop_while is_assume (CCVector.to_list stmts)
+            in
+            if List.exists is_assume rest then
+              failwith "block has non-contiguous assumes";
```
agle: I don't think it should be neccessary to move the guard back to the DSA intermediate block, the phi node is always before the guard? I might be missing the reasoning here though. This kind of thing is tricky to get correct wrt the most simple constant time preservation proof (leaking conditions) so i was intending to avoid.

katrinafyi: Yeah I think I wasn't sure how much "lookahead" the a nondeterministic branch was allowed to have. I will remove.

agle: Yeah i decided to just try make the interpreter able to fully explore any path and rollback this time- boogie or a symbolic executor would handle it- its part of the semantics so we should just support it :)

lib/transforms/dsa.ml
katrinafyi: Oopsie - https://github.com/agle/bincaml/pull/138

PR #137, opened 2026-04-24
tests: add `tree-sitter highlight` to catch invalid node type

katrinafyi:
if tree-sitter is available, it will raise an error if the highlights file has outdated node references
```
Error in query file ./queries/highlights.scm

Caused by:
    Query error at 93:2. Invalid node type blah
```
tree-sitter tests still don't run in CI, but it means that if someone is working on the tree-sitter, they'll probably be able to see this error.

In future, this should prevent https://github.com/agle/bincaml/pull/134

### DIFF
--- a/tree-sitter/dune
+++ b/tree-sitter/dune
@@ -1,11 +1,21 @@
 ; checks tree-sitter is valid and unambiguous
 
 (rule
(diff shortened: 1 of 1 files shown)


JTrenerry (approved):
Would be helpful

PR #141, opened 2026-04-28
fix unstable attrib order in boogie backend

agle:
use bindings instead of to_list


katrinafyi:
why does this fix it?

agle:
CCMap.Make.to_list does not define an order, and different versions of the library return a different order (it was changed to choose the more performant fold direction). `bindings` defines the order according according to the map order.

PR #142, opened 2026-04-29
Strongest Postcondition

mira-alford:
Strong post-condition ensures clause generation/propagation. Yay!

Integrates into the procedure summaries transform nicely. Seems to work out on the simple test case I've written, and also some memory encoding tests ill probably have in a PR next week.

Some things are just sent to top like loads, not sure how I would write a transfer function for them, but this is enough to get plenty of interesting examples passing.


lib/analysis/dune
```diff
@@ -18,7 +18,9 @@
   tnum_wint_reduced_product
   known_bits
   wp_dual
-  sva)
+  sp
+  sva
+  linear_const)
```
b-paul: linear const is already there lol (merge mistake i assume)

lib/analysis/intra_analysis.ml
```diff
@@ -166,6 +166,7 @@ module Forwards (D : IntraDomain) = struct
       ?(widening_set = Graph.ChaoticIteration.Predicate (fun _ -> false))
       ?(widening_delay = 0) p =
     Trace_core.with_span ~__FILE__ ~__LINE__ D.name (fun _ ->
+        let g = Procedure.graph p in
```
b-paul: unused ?

lib/analysis/sp.ml
```diff
@@ -0,0 +1,263 @@
...
+        | _ -> Keep)
+    in
+    rw % rw
+
+  let unwrap p = Option.get_or ~default:e_true p
+  let show = Option.map_or ~default:"" BasilExpr.to_string
+  let equal = Option.equal BasilExpr.equal
+  let compare = Option.compare BasilExpr.compare
+  let pretty = BasilExpr.pretty % unwrap
+  let top = None
```
b-paul: I'd think that t could just be BasilExpr.t with top being e_true, that way join top top = top and also much of the logic would be simplified

lib/analysis/sp.ml
```diff
@@ -0,0 +1,263 @@
...
+    let o =
+      match (a, b) with
+      | None, None -> Some e_true
+      | None, x | x, None -> x
+      | Some a, Some b -> Some (BasilExpr.applyintrin ~op:`OR [ a; b ])
+    in
+    o
+
+  let leq a b = failwith "leq not implemented"
+  let widening a b = bottom
```
b-paul: widening should approximate join so i think should be top, not bottom

lib/analysis/sp.ml
```diff
@@ -0,0 +1,263 @@
...
+            s
+            |> substitute_var_names (StringMap.to_list args)
+            |> substitute_var_names
+                 (StringMap.to_list lhs
+                 |> List.map (fun (i, v) -> (i, BasilExpr.rvar v)))
+          in
+          let r =
+            List.fold_left
+              (fun a b -> conjunction [ a; sub b ])
+              e_true (S.ensures procid)
```
b-paul: it would probably be better not to have this as a fold and instead just have the ensures clause appended to the singleton list [a] before calling conjunction, so that there aren't layers of two element conjunctions (applies to other conjunction/join folds)

lib/analysis/sp.ml
```diff
@@ -0,0 +1,263 @@
...
+          (fun (v', e) -> Option.return_if (String.equal (Var.name v) v') e)
+          a)
+      p
+    |> simplify
+
+  let conjunction ls = BasilExpr.applyintrin ~op:`AND ls
+
+  let tf_assigns p a =
+    a
+    |> List.map (function a, b -> BasilExpr.binexp ~op:`EQ (BasilExpr.rvar a) b)
```
b-paul, line 120: I'm worried about this creating predicates that are too large... but i'm not too sure how to reduce the size other than not storing information about variables that aren't relevant to the output

lib/analysis/wp_dual.ml
```diff
@@ -1,15 +1,16 @@
 (* The wp dual domain for precondition inference *)
+(* Assumes SSA form *)
 
 open Lang
 open Common
 open Expr
 
-module type FunctionSummaryAnnotation = sig
+module type Functionnnotation = sig
```
b-paul: I think you forgot an A oops

lib/lang/expr.ml
```diff
@@ -505,7 +505,7 @@ module BasilExpr = struct
         let binding =
           fill (text " ")
             (List.map (fun v -> bracket "(" (Var.pretty v) ")") bound_vars)
-          ^+ sep ^+ a ^ bracket "(" b ")"
```
b-paul, line 508: wouldn't this drop attributes from dumped il files?

lib/transforms/function_summaries.ml
```diff
@@ -19,6 +19,13 @@ module type FunctionSummaryAnnotation = sig
   val ensures : ID.t -> Expr.BasilExpr.t list
 end
 
+module type FunctionAnnotation = sig
+  include FunctionSummaryAnnotation
```
b-paul: Can FunctionSummaryAnnotation be replaced with this module

lib/transforms/function_summaries.ml
```diff
@@ -76,14 +83,25 @@ let wp_dual_requires (module S : FunctionSummaryAnnotation)
   Analysis.A.M.find_opt Procedure.Vert.Entry result
   |> Option.map Domain.to_pred |> Option.to_list
 
+let sp_ensures (module S : FunctionAnnotation) (proc : Program.proc) =
+  let module Domain = Sp.Domain (S) in
+  let module Analysis = Intra_analysis.Forwards (Domain) in
```
b-paul, line 83: Do you know if it's possible to use the DFG solver? that would be cool (needs to handle asserts/assumes though)

lib/transforms/livevars.ml
```diff
@@ -86,10 +86,9 @@ let%expect_test _ =
   print_endline (to_string e2);
   [%expect
     {|
-    forall (v1:bv1) :: (eq(v2:bv1,
-      forall (v2:bv1) :: (booland(v1:bv1, v2:bv1, v3:bv1))))
-    forall (v1:bv1) :: (eq(0x16:bv5,
-      forall (v2:bv1) :: (booland(v1:bv1, v2:bv1, 0x16:bv5))))
+    forall (v1:bv1) ::(eq(v2:bv1, forall (v2:bv1) ::(booland(v1:bv1, v2:bv1, v3:bv1))))
```
b-paul: oh no a space got dropped between the separator

b-paul:
it's probably important that some bound on the size of predicates is put in so that they don't explode in size

mira-alford:
> it's probably important that some bound on the size of predicates is put in so that they don't explode in size

This was worrying me greatly, given how large such a small example got. I think Bounding it is probably the best idea for now.

lib/analysis/sp.ml
mira-alford: Idk why it was happening but the issue i ran into with this (it was my original solution) was that init get called once for every node, causing it to join with true, and as join is OR that just ended up killing any predicates. I added the extra optional value so that the entrypoint to the procedure could be initialised to "nothing" but then future inits are discarded. Any idea about a better way to handle this?

lib/analysis/wp_dual.ml
mira-alford: I swear i fixed this end of yesterday lol. Maybe forgot to push it :)

lib/lang/expr.ml
mira-alford: No, jackson and i were very confused by this. They were actually getting dumped twice for some reason, once before the :: and once after. We found it because the lsp was exploding on the printed output because it didn't match the grammar.

b-paul: terrifying...

lib/analysis/sp.ml
b-paul: joining with top should be top, i think you want to initialise the entry block to top (or the current requires clause even) and have every other block initialise to bottom, then the solver will propagate information forwards

mira-alford: Yeah, that was the goal. Didn't seem like there was any good way to do that though?

b-paul: true... the place that that's configured is at the call to ocamlgraph's chaotic iteration [here](https://github.com/agle/bincaml/blob/main/lib/analysis/intra_analysis.ml#L172) and this api doesn't support dynamically choosing the initial value based on the block... maybe making the vertex a param to D.init could be worth it?

lib/transforms/livevars.ml
JTrenerry: https://github.com/agle/bincaml/pull/142/changes/BASE..482f8262560e4c4195b00daef8b06233757465fa#diff-580350c5770973c441494d583e7eeb7d463b71b5046b4fba2baa46d2f316d869L508

Yeah, change this to be ^+ and not ^ i guess, I don't know what ^+ is but I assume its cat with a space

mira-alford:
@b-paul would love a quick re review whenever your able. Should hopefully be good to merge now.

b-paul (approved):
looks good yay

PR #145, opened 2026-05-08
IDE ssi solver with abi

b-paul:
Assume R19 to R29 and R31 are preserved across external function calls to get better simplification


JTrenerry (approved):

JTrenerry:
string matching = bad, so maybe eventually something more sound could be used? (i.e. call_preserved flag or something but that would be just string matching to begin with lol)

agle:
I've unmerged this, going down this road ends up in the mess basil had. We need a systematic approach to abi not hacks in each individual transform/analysis

PR #148, opened 2026-05-12
repeated ssa / phi node support

mira-alford:
Adds support to SSA transform for repeated application of SSA, and as a requirement of that also that SSA now supports pre-existing phi nodes in the IR. Adds 3 corresponding tests.


b-paul (approved):
lgtm 👍 👍

PR #149, opened 2026-05-14
types cleanup

agle:
- explicitly represent types in the expression ast
- remove expr from attribute type and move triggers to expr ast
- add memory set/ get to grammar so memory encoding output is parseable
- fix loader adding nop instructions in return blocks


lib/transforms/livevars.ml
```diff
@@ -86,10 +86,10 @@ let%expect_test _ =
...
   [%expect
     {|
-    forall (v1:bv1) :: (eq(v2:bv1,
-      forall (v2:bv1) :: (booland(v1:bv1, v2:bv1, v3:bv1))))
-    forall (v1:bv1) :: (eq(0x16:bv5,
-      forall (v2:bv1) :: (booland(v1:bv1, v2:bv1, 0x16:bv5))))
+    forall  (v1:bv1) :: (eq(v2:bv1,
+      forall  (v2:bv1) :: (booland(v1:bv1, v2:bv1, v3:bv1))))
+    forall  (v1:bv1) :: (eq(0x16:bv5,
+      forall  (v2:bv1) :: (booland(v1:bv1, v2:bv1, 0x16:bv5))))
```
JTrenerry: All the test cases seem to have an extra space, there was one above with an exists doing it as well.

lib/analysis/sp.ml
```diff
@@ -246,7 +249,7 @@ proc @branching(a:bv64) -> (b:bv64) [
   |> IntraDomain.to_pred |> BasilExpr.to_string |> print_endline;
   [%expect
     {|
-    exists (a_1:bv64) (a_3:bv64) (x_2:bv64) (a_2:bv64) (x_1:bv64) (x_3:bv64) :: (boolor(booland(boolor(booland(boolor(booland(eq(a_1:bv64,
+    exists  (a_1:bv64) (a_3:bv64) (x_2:bv64) (a_2:bv64) (x_1:bv64) (x_3:bv64) :: (boolor(booland(boolor(booland(boolor(booland(eq(a_1:bv64,
```
JTrenerry: space?

lib/backends/boogie.ml
```diff
@@ -211,20 +210,17 @@ and pretty_attribute_map (key : string) (a : Program.e Attrib.attrib_map) =
...
+  |> List.map (fun attrib -> bracket "{" (append_l ~sep:(text ", ") attrib) "}")
+  |> append_sp
 (* Option.map (Attrib.attrib_pretty Expr.BasilExpr.pretty) attrib |> Option.get_or ~default:(text "MAGIC") *)
 
-and pretty_binding_expr ?(attrib : Program.e Attrib.t option) bound in_body =
+and pretty_binding_expr triggers bound in_body =
   let open Containers_pp in
   pretty_call_args_no_brackets (List.map pretty_variable_typed bound)
-  ^+ text "::" ^+ newline_or_spaces 0 ^ pretty_triggers attrib
+  ^+ text "::" ^+ newline_or_spaces 0 ^ pretty_triggers triggers
```
JTrenerry, line 223: Assume it is from empty pretty_call_args_no_brackets?

katrinafyi:
Re the lint-doc failure, it looks like there's a warning which is treated-as-error and this isn't a warning due to doc comments we've written.
```
Error: Failed to compile expansion for include module type of struct include identifier(root(Bincaml_util__ID).ID,false) end Unresolved module type path identifier(root(Bincaml_util__ID).ID_Pub,false) (Lookup failure (module type): root(Bincaml_util__ID).ID_Pub)
File "bincaml_util__ID.cmt":
Error: Failed to compile expansion for module type expression identifier(root(Bincaml_util__ID).ID_Pub,false) Unresolved module type path identifier(root(Bincaml_util__ID).ID_Pub,false) (Lookup failure (module type): root(Bincaml_util__ID).ID_Pub)
ERROR: Warnings have been generated.
```
I think the goal of the lint-doc rule was to catch broken `{!Thing}` links which we manually write, but I think we can just delete the lint-doc rule. We already have the separate `docs / Build odoc` task.

Issue #130, opened 2026-04-17
Encode and check transform dependencies


agle:
Its currently unclear to a user what transforms are neccessary or how to appropriately order them. This issue is to address that by encoding invariants on the IR which transforms can require and ensure. The set of invariants can be simply a set or list attched to each pass definition.

- define adt for list of invariants: (required for many transforms) SSA, (required for boogie output) DSA, (required for ssa) Params
- specify contract of each pass considering:
   - some passes may need to require an invariant to be not present (e.g. not in ssa )
   - passes will break/remove an invariant, this will need to be encoded
   - property checks can ensure a property without performing a transform
- encode the current set of invariants a program satisfies in a state fed through passes.ml, this will require wrapping the program. We can maybe factor out some more of the code handling script evaluation into passes.ml
    - do the invariants apply per-procedure or to the whole program?
- before each pass check the property holds and fail or produce an error message when it isnt satisfied. Failure is likely not desirable as in some cases the user will know better than the dependency checker.

PR #150, opened 2026-05-18
Boogie Backend Phi Nodes

mira-alford:
Got around to supporting phi nodes in the boogie backend.

Much like everything in the boogie prepass it kind of mangles the IR, but it seems to generate the correct assigns corresponding to phi nodes so yay! If only boogie had explicit phi nodes...

Also added 2 test cases. One for basic boogie gen, and then one more complicated one with a few memory encoding examples.

### DIFF
--- a/test/cram/dune
+++ b/test/cram/dune
@@ -9,7 +9,14 @@
   concat.il))
 
 (cram
- (applies_to * \ expr_smt malloc_free memory_safety memory_interproc)
+ (applies_to
+  *
+  \
+  expr_smt
+  malloc_free
+  memory_safety
+  memory_interproc
+  memory_phis)
  (package bincaml)
  (deps
   loop_dfg.sexp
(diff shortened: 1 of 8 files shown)


test/cram/dune
```diff
@@ -9,7 +9,14 @@
...
 
 (cram
- (applies_to * \ expr_smt malloc_free memory_safety memory_interproc)
+ (applies_to
+  *
+  \
+  expr_smt
+  malloc_free
+  memory_safety
+  memory_interproc
```
mira-alford, line 18: This is kind of getting unwieldy every time i add a test that uses boogie... is there a better way to do this? @agle

agle:
@katrinafyi already implemented this as a discrete transform pass https://github.com/agle/bincaml/pull/129, the backend should just fail if phi nodes are present. We should implement https://github.com/agle/bincaml/issues/130 to address confusion like this. I'm also not sure this implementation is correct i think there can technically be a situation where in the presence of loops dependencies are introduced between the introduced assignments as theyre not simultaneous anymore, (e.g. a phi node swaps two variables).

PR #151, opened 2026-05-18
add codeowners

agle:
Requires myself or Kait's review of every PR

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,10 @@
+
+* @agle @katrinafyi
+
+/lib/*.ml @agle
+/lib/lang @agle
+
+/lsp @katrinafyi
+/tree-sitter @katrinafyi
+/nix @katrinafyi
+


CODEOWNERS
```diff
@@ -0,0 +1,10 @@
+
+* @agle @katrinafyi
+
+/lib/*.ml @agle
+/lib/lang @agle
```
katrinafyi, line 5: Just double-checking that the subfolder rules work in the way we want, so it's not additive with the parent folder, right?

katrinafyi (approved):

CODEOWNERS
agle: Later rules are are more restrictive than earlier ones so I think this requires me to look at anything under lib/lang https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file

agle:
Good to know CI just dislikes me personally

PR #152, opened 2026-05-18
Common sub-expression elimination pass

agle:
Uses hashconsing to identify common sub expression and create a fresh variable to substitute, as soon as all the expressions' dependencies are defined. We do the substitution asap in a preorder traversal of the dominator tree, hopefully thats safe alone, since we have proper ssa form, it should maximise the substitutions of each expr. 

Its not super efficient but seems good enough for cntlm.

image

The nonobvious question is whether this is a good optimisation; there is a trade-off between storage and computation and this pushes towards storage which might make some analyses slower. 

It might be interesting with the SVA and type inference though, as it will pull the stack addresses into a distinct variable that gets reused throughout the program that may be easier on SMT as well as providing something we can easily make pointer-typed.

image

There's some extra refactoring bug-fixes and debugging printers  tests to do with the hashconsed expressions. There's a bit of a trap there where the source-code locations embedded in the expressions make two identical looking expressions unique if you construct a hashconsed expr with the obvious `BasilExpr.cata HashedExpr.fix expr`, you need to do `BasilExpr.cata (AbstractExpr.drop_attrib %> HashedExpr.fix) expr` or `HashedExpr.of_expr`. `BasilExpr.show_abstract` can print an abstract expr in detail using a derived pp to hopefully make this more visible.

I haven't done a semantic test of it, the interpreter-based fuzzing or translation validation would be good.

image


lib/lang/expr.ml
```diff
@@ -97,31 +106,58 @@ module AbstractExpr = struct
...
+        BinaryExpr { op = op2; arg1 = arg11; arg2 = arg12 } ) ->
+        Equal.triple eq_binary eq_expr eq_expr (op, arg1, arg2)
+          (op2, arg11, arg12)
+    | ApplyIntrin _, ApplyIntrin _ -> true
+    | ApplyFun _, ApplyFun _ -> true
+    | Lambda _, Lambda _ -> true
+    | Let _, Let _ -> true
+    | _ -> false
+    *)
+
```
katrinafyi: is the new code functionally identical to old code or is there some change in behaviour? new code below
```
          AbstractExpr.equal equal_const Var.equal equal_unary equal_binary
            equal_intrin Types.equal equal i' j'
```

```suggestion
```

lib/lang/expr.ml
```diff
@@ -962,6 +977,72 @@ module BasilExpr = struct
...
+        let he = hash i = hash j in
+        match (i, j) with
+        | E i', E j' ->
+            let e =
+              AbstractExpr.equal equal_const Var.equal equal_unary equal_binary
+                equal_intrin Types.equal Fix.HashCons.equal
+                (AbstractExpr.drop_attrib i')
+                (AbstractExpr.drop_attrib j')
+            in
+            if e then if not he then raise (EQErr (i, j));
```
katrinafyi: what's this error for?

lib/lang/ops.ml
```diff
@@ -523,6 +523,8 @@ module AllOps = struct
     | `IfThen -> "case"
     | `Cases -> "match"
 
+  let hash a = to_string a |> String.hash
```
katrinafyi: compared to the hand-coded functions for expr, this is amusingly inefficient.

lib/lang/procedure.ml
```diff
@@ -694,3 +700,38 @@ let pretty show_lvar show_var show_expr p =
     | None -> nil
   in
   header ^ spec ^ newline ^ blocks
+
+module BlockGraph = struct
+  (* simpler graph of block dependency *)
```
katrinafyi:
```suggestion
(** A simpler graph of block dependency *)
module BlockGraph = struct
```

lib/transforms/cse_elim.ml
```diff
@@ -0,0 +1,331 @@
...
+  let full =
+    Expr.BasilExpr.HashExprFix.unfix e
+    |> Expr.BasilExpr.show_abstract (fun f e ->
+        Format.fprintf f "%d" @@ Fix.HashCons.id e)
+  in
+  let t =
+    Expr.BasilExpr.ExprHashCons.cata Expr.BasilExpr.fix e
+    |> Expr.BasilExpr.to_string_annot
+  in
+  Printf.sprintf "%d:%d:%d=%s==%s" i h ihash t full
```
katrinafyi: should these be inside the hash expr module? actually, should the hashexpr module be its own module next to expr.ml? i think it would make it neater

lib/transforms/cse_elim.ml
```diff
@@ -0,0 +1,331 @@
...
+
+let compl e =
```
katrinafyi: what is this short for? rename or add comment

lib/transforms/cse_elim.ml
```diff
@@ -0,0 +1,331 @@
...
+
+let tf_stmt st ~thresh p s =
+  let open Stmt in
+  let new_substs st =
+    ESet.diff st.allowed (ESet.of_iter @@ EMap.keys st.avail)
+    |> ESet.filter (fun v ->
+        VarSet.is_empty
+        @@ VarSet.diff (BasilExpr.ExprHashCons.free_vars v) st.defined)
+    |> ESet.to_iter
+    |> Iter.uniq ~eq:Expr.BasilExpr.ExprHashCons.equal
```
katrinafyi: why uniquify something which was just a set?

lib/transforms/cse_elim.ml
```diff
@@ -0,0 +1,331 @@
...
+    |> EMap.keys |> ESet.of_iter
+  in
+  subst
+
+let block_defs p =
+  Procedure.iter_blocks p
+  |> Iter.flat_map (fun (id, b) ->
+      Block.assigned_vars_iter b |> Iter.map (fun v -> (v, b)))
+
+let do_proc ~thresh p =
```
katrinafyi: ###  Thresh
<a href="https://www.leagueoflegends.com/en-us/champions/thresh/">
Image
</a>

That is to say, I think we can just use the full word.

lib/transforms/cse_elim.ml
```diff
@@ -0,0 +1,331 @@
...
+
+    [
+       block blah [
+         var A:bv64 := bvnot(0x1:bv64);
+         var B:bv64 := bvnot(0x1:bv64);
+         var C:bv64 := bvnot(0x1:bv64);
+         var D:bv64 := bvnot(0x1:bv64);
+         return;
+       ]
+    ]proc <proc>()  -> () {  }
```
katrinafyi: what happened here?

lib/transforms/cse_elim.ml
```diff
@@ -0,0 +1,331 @@
...
+  let eq =
+    List.init 5 (fun _ ->
+        BasilExpr.unexp ~op:`BVNOT
+          (BasilExpr.bvconst (Bitvec.of_int ~size:32 5))
+        |> BasilExpr.ExprHashCons.of_expr
+        |> fun id ->
+        Printf.sprintf "%d %d" (Fix.HashCons.id id) (Fix.HashCons.hash id))
+    |> List.to_string ~sep:", " Fun.id
+  in
+  print_endline eq;
```
katrinafyi: these seem like smaller more basic tests. can these move into a separate test and be more descriptive? is  the `\nnn` intentional?

lib/transforms/cse_elim.ml
```diff
@@ -0,0 +1,331 @@
...
+let%expect_test "hash" =
+  let a0 = Loader.Loadir.parse_expr_string "bvnot(1:bv64)" in
+  let a = BasilExpr.ExprHashCons.of_expr a0 in
+  let b0 = Loader.Loadir.parse_expr_string "bvnot(0x1:bv64)" in
+  let b = BasilExpr.ExprHashCons.of_expr b0 in
+  Printf.printf "%s =\n%s\n" (show_he a) (show_he b);
+  if BasilExpr.ExprHashCons.equal a b then print_endline "equal"
+  else print_endline "not_equal";
+  if BasilExpr.equal a0 b0 then print_endline "o equal"
+  else print_endline "o not_equal";
```
katrinafyi:
```suggestion
  Printf.printf "equal after hash-cons: %B\n" (BasilExpr.ExprHashCons.equal a b);
  Printf.printf "equal before hash-cons: %B\n" (BasilExpr.equal a0 b0);
```

lib/transforms/cse_elim.ml
```diff
@@ -0,0 +1,331 @@
...
+    Expr.BasilExpr.ExprHashCons.cata Expr.BasilExpr.fix e
+    |> Expr.BasilExpr.to_string_annot
+  in
+  Printf.sprintf "%d:%d:%d=%s==%s" i h ihash t full
+
+let compl e =
+  AbstractExpr.(
+    match e with RVar _ -> 1 | Constant _ -> 1 | _ -> fold ( + ) 1 e)
+
+let visit_block st ~thresh s =
```
katrinafyi: rename: this is doing some counting of some kind

lib/transforms/cse_elim.ml
```diff
@@ -0,0 +1,331 @@
...
+    in
+    let x = AbstractExpr.map snd e |> AbstractExpr.fold ESet.union ESet.empty in
+    let x =
+      if ESet.mem hc st.allowed && not (EMap.mem hc st.avail) then ESet.add hc x
+      else x
+    in
+    (hc, x)
+  in
+  let to_add = BasilExpr.cata alg e in
+  snd to_add
```
katrinafyi: unused?

lib/transforms/cse_elim.ml
```diff
@@ -0,0 +1,331 @@
...
+  in
+  let substs_after =
+    new_after |> EMap.of_iter |> EMap.union (fun k v1 v2 -> Some v2) st.avail
+  in
+  let substs_after' =
+    Iter.map (fun (e, v) -> Instr_Assign [ (v, to_basil e) ]) new_after
+  in
+  let st = { st with avail = substs_after } in
+  ( st,
+    Iter.append substs_bef' @@ Iter.append (Iter.singleton stmt') substs_after'
```
katrinafyi:
```suggestion
    Iter.append substs_bef' @@ Iter.cons stmt' substs_after'
```

lib/transforms/cse_elim.ml
```diff
@@ -0,0 +1,331 @@
...
+  let substs_bef' =
+    Iter.map (fun (e, v) -> Instr_Assign [ (v, to_basil e) ]) new_bef
+  in
+  let substs_after =
+    new_after |> EMap.of_iter |> EMap.union (fun k v1 v2 -> Some v2) st.avail
+  in
+  let substs_after' =
+    Iter.map (fun (e, v) -> Instr_Assign [ (v, to_basil e) ]) new_after
+  in
+  let st = { st with avail = substs_after } in
```
katrinafyi: can we have comments to convey the main ideas of this long sequence?

katrinafyi (commented):
i can't follow the cse logic. can we have renames or comments?

these names are not constrained to some visitor interface, there is no need to make names so generic like visit_block, vstmt, vexpr, tf_stmt, and probably others.

lib/lang/expr.ml
agle: It should be but I was considering ignoring attrib

lib/transforms/cse_elim.ml
agle: 😡️

JTrenerry:
For type inference, are you expecting improvements (speed or type inference wise) or just well type-ness coming from pointers (ptradd returning bvs)

agle:
I don't remember what the case was where the ptr type inference didn't work, but I hope it would infer a better type for the pointers.

katrinafyi (approved):

PR #154, opened 2026-05-19
Fix linear const termination bug

b-paul:
Linear functions didn't have a canonical representation so there could be multiple ways to write top, and the solver assumed there was a unique repr (using a <= b iff a join b == b) so this fixes that

### DIFF
--- a/lib/analysis/linear_const.ml
+++ b/lib/analysis/linear_const.ml
@@ -81,6 +81,19 @@ module LF = struct
       Some (Value.V j)
     else None
 
+  let make_linear a b =
+    match () with
+    | _ when Z.equal Z.one (Bitvec.value a) && Z.equal Z.zero (Bitvec.value b)
+      ->
+        IdEdge
+    | _ -> Linear (a, b)
+
+  let make_join a b c =
+    match c with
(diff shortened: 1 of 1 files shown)


lib/analysis/linear_const.ml
```diff
@@ -81,6 +81,19 @@ module LF = struct
       Some (Value.V j)
     else None
 
+  let make_linear a b =
+    match () with
+    | _ when Z.equal Z.one (Bitvec.value a) && Z.equal Z.zero (Bitvec.value b)
```
katrinafyi: match on the bool predicate or use an ITE :)

agle (approved):

lib/analysis/linear_const.ml
katrinafyi: am I right in guessing that this case won't be seen anymore after canonicalisation? is it worth removing or assert false?

katrinafyi (approved):
looks reasonable. can you add a regression test as well? otherwise lgtm

lib/analysis/linear_const.ml
b-paul: true it's probably worth making sure those cases don't get hit

PR #156, opened 2026-05-19
Enable and fix some warnings

katrinafyi:


dune
```diff
@@ -14,3 +14,17 @@
   (package bincaml_lsp)
   bincaml_lsp.install)
  (package bincaml_lsp))
+
+(env
+ ; in dev, all warnings are non-fatal.
+ (dev
+  (flags
+   (:standard -warn-error -A -w -27-9-33)))
```
agle: Can we keep unused-open (-33) , could be a separate PR to re-enable and fix. 

Also in these places is it possible to use the name of the warning rather than the number? E.g. `-unused-var-strict -missing-record-field-pattern`

lib/lang/procedure.ml
```diff
@@ -215,8 +215,11 @@ end = struct
   let map_formal_out_params f p =
     { p with formal_out_params = f p.formal_out_params }
 
-  let set_formal_in_params f p = map_formal_in_params (fun _ -> f) p
-  let set_formal_out_params f p = map_formal_in_params (fun _ -> f) p
+  let[@warning "-32"] set_formal_in_params f p =
+    map_formal_in_params (fun _ -> f) p
+
+  let[@warning "-32"] set_formal_out_params f p =
+    map_formal_in_params (fun _ -> f) p
```
agle: This probably means its missing from the mli file and should be added

lib/analysis/gamma_domain.ml
```diff
@@ -14,7 +15,7 @@ module GammaSet = struct
   end)
 
   let name = "Gamma set"
-  let pp fmt v = Format.pp_print_string fmt (show v)
+  let[@warning "-32"] pp fmt v = Format.pp_print_string fmt (show v)
```
agle: Should we instead add pp to the interface? We generally want these for deriving printers right?

dune
katrinafyi: I'll leave unused-open to another PR because it's just so extensive. Tbh, maybe it's okay to leave that warning off because it might fire too often if we import but don't use Common, for example.

katrinafyi (commented):
All comments actioned aside from one comment I've replied to. Thanks!

agle (approved):

PR #159, opened 2026-05-21
Examples submodule

agle:
Moves examples to a submodule so we can safely include more without blowing up the repo size.


JTrenerry:
To do stuff to the examples git directory, is it just cd into it and then git commands will work there?

Also how do we go about cram testing now? Should we avoid using ones in examples (tbf reasonable to be doing that before) or do we just assume anyone running tests has it? or is there a condition to check if its there and to not run it?

agle:
CI checks out the submodule so cram tests should use them, generally its better to put the IL files in the submodule rather than the tests dir I think. With the submodule we can worry less about adding more and larger files but I guess should still be a *little* careful not to make it a huge repo that takes forever to clone.

Git submodules can be a bit finnicky to work with but generally yeah it functions as a separate repo attached to that directory; you can just add files and commit and push to main in the submodule repo it should be fine. The submodule itself points to a specific commit of repo though, so from the bincaml root you have to git add the directory so it points to the new commit.

JTrenerry:
Yeah submodules terrify me :(

readme.md
```diff
@@ -4,6 +4,21 @@ lightweight binary exploration verification project implemented in ocaml.
...
+
+```
+git clone --recurse-submodules https://github.com/agle/bincaml.git --shallow-submodules
+```
+
+If you have already cloned you can fetch the submodule with:
+
+```
+git submodule init
+git submodule update
```
katrinafyi:
```suggestion
git submodule update --init --depth 1
```

to combine the two commands and check out shallowly, lining up with the `clone --shallow-submodules
`.

.github/workflows/test.yml
```diff
@@ -54,6 +56,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+        with:
+          submodules: true
```
katrinafyi, line 60: i assume the submodule isn't needed atm for deploy-doc.yml, but do you want to add it anyway for consistency and to avoid future surprise?

katrinafyi (approved):
what is the workflow for updating the external UQ-PAC/bincaml-test repository? if a bincaml pr needs new test files, those will need to be pushed to bincaml-test before it can be referenced via the bincaml submodule. how does it work when two PRs try to add .il files at the same time? will that appear as an opaque conflict on the submodule hash or does git do something nicer?

after this is merged, users should be aware that they will need to init submodules, otherwise they will continue using their local and untracked examples folder.

.github/workflows/test.yml
agle: No the doc build shouldn't have to depend on the examples and its faster to build without checking it out.

agle:
> what is the workflow for updating the external UQ-PAC/bincaml-test repository? if a bincaml pr needs new test files, those will need to be pushed to bincaml-test before it can be referenced via the bincaml submodule. how does it work when two PRs try to add .il files at the same time? will that appear as an opaque conflict on the submodule hash or does git do something nicer?

1. to update the examples

```
cd examples && git add . && git commit -m 'add example' && git push
cd .. && git add examples && git commit -m 'update examples'
```

2. Merge conflicts will look like the below but the only possible conflict situation is that one of the branches is behind main (as updating the examples and resolving conflicts should have already happened in the other repository) so there's an obvious resolution.

```
diff --git a/examples b/examples
index 8a8cd517..b9b6da4f 160000
--- a/examples
+++ b/examples
@@ -1 +1 @@
-Subproject commit 8a8cd5178397c70e1d62ab046da659ba8370be7e
+Subproject commit b9b6da4f2062474e63ad0080f231fd2599491453
```

With the resolution being

```
cd examples && git pull && cd .. && git add examples && git commit -m 'update examples' && git push
```

> after this is merged, users should be aware that they will need to init submodules, otherwise they will continue using their local and untracked examples folder.

There will be a merge/rebase conflict when it tries to checkout the submodule directory

katrinafyi:
> There will be a merge/rebase conflict when it tries to checkout the submodule directory

There wasn't for me.

Copy the rest of the information into a docs.

agle:
Its really just submodules 101 but I guess it can't hurt

katrinafyi:
Submodule users might want this if they want to push to the submodule repo:
```
[url "git@github.com:"]
  pushInsteadOf = "https://github.com/"
```
in `~/.gitconfig`.

PR #160, opened 2026-05-22
basic gtirb loader

agle:
Initial skeleton to load gtirb files

- load intraprocedural CFG and build basic block structure pre/post conditions of PC change on each block.
- inject assert false for each opcode annotated by opcode and disassembly.

The disassembly subprocess setup should hopefully be temporary, it looks for any version of llvm-mc in path, ideally we replace it with capstone or kait's assembler if it becomes stable.


katrinafyi:
What's happened to the git history? 

in any case, I approve in principle and I look forward to a well architected frontend. Also thanks for removing capstone :)

agle:
I must have messed up the history when I rebased out capstone

agle:
It needs more testing, and I'm a bit suspect of the inter-procedural handling https://github.com/agle/bincaml/pull/160/changes#diff-88ba6b3ec962119dc4eb142e21586381a4ceb7f2a6d4cbdee205792f80df0fbdR73, I think it needs another check that the call target doesn't have successors. The other issue is that the PC assignments will break across calls currently because we don't ensure the return address of the call.

agle:
> The other issue is that the PC assignments will break across calls currently because we don't ensure the return address of the call.

I believe this will be simplest with an ABI-specific pass as the return address is dependent on the call address, which we have access to through R31.  We could attempt to invent a fake register for this purpose but I don't think its worth it.

lib/gtirb/disas.ml
```diff
@@ -0,0 +1,40 @@
+open Lang
+open Common
+
+(** Subprocess wrapper around llvm-mc *)
+let llvm_mc_bin =
+  lazy
+    (Iter.cons "llvm-mc"
+       (Iter.int_range_dec ~start:19 ~stop:15
```
katrinafyi, line 8: future proofing

```suggestion
       (Iter.int_range_dec ~start:30 ~stop:15
```

lib/gtirb/gfir.ml
```diff
@@ -0,0 +1,218 @@
...
+module AD = Load_auxdata.Loaders
+(** Auxdata parsers *)
+
+open Conf
+
+(** {2 GTIRB Frontend Intermediate Representation} *)
+
+(** edges just copied directly from gtirb protobuf CFG *)
+module Edge = struct
+  type edge_type =
```
katrinafyi: if these are copied from the protobuf can we use a type equality constraint like `type edge_type = CFG.Proto.Edge = ... = ...|...`

lib/gtirb/gfir.ml
```diff
@@ -0,0 +1,218 @@
...
+    List.map (fun (b : Gtirb.block) -> (b.uuid, b)) blocks |> UUIDMap.of_list
+  in
+  let auxdata =
+    m.aux_data |> StringMap.of_list |> StringMap.filter_map (fun _ v -> v)
+  in
+  let or_error default = function
+    | Ok e -> e
+    | Error msg ->
+        Logs.warn (fun m -> m "load:gtirb %s" msg);
+        default
```
katrinafyi: this function doesn't need to be an inner function

lib/gtirb/gfir.ml
```diff
@@ -0,0 +1,218 @@
...
+                UUIDMap.get id all_blocks |> fun a ->
+                Option.bind a (fun b -> Some (id, b)))
+            |> UUIDMap.of_iter )
+      | None ->
+          Logs.warn (fun m ->
+              m "No entry blocks for proc: %s %s" (UUID.show uuid) name);
+          (UUIDSet.empty, UUIDMap.empty)
+    in
+    let id = Lang.Program.declare_name ("@" ^ name) prog in
+    { uuid; name; id; entries; blocks; code_blocks; cfg = G.empty }
```
katrinafyi: now does this one?

lib/gtirb/gfir.ml
```diff
@@ -0,0 +1,218 @@
...
+        |> List.filter (fun { source_uuid; _ } ->
+            block_member_of ~proc:p.uuid (UUID.of_bytes source_uuid))
+        |> List.map (fun { source_uuid; label; target_uuid } ->
+            ( conv_vert @@ UUID.of_bytes source_uuid,
+              E.of_gtirb_label label,
+              conv_vert @@ UUID.of_bytes target_uuid )))
+    in
+    let cfg = List.fold_left (fun cfg e -> G.add_edge_e cfg e) p.cfg edges in
+    { p with cfg }
+  in
```
katrinafyi: does this one?

lib/gtirb/gfir_to_bincaml.ml
```diff
@@ -0,0 +1,280 @@
...
+open Ocaml_protoc_plugin
+open IR.Gtirb.Proto
+open ByteInterval.Gtirb.Proto
+open Module.Gtirb.Proto
+open Section.Gtirb.Proto
+open CFG.Gtirb.Proto
+open Load_auxdata
+module UUIDMap = UUIDMap
+module UUIDSet = UUIDSet
+open Gfir
```
katrinafyi, line 12: can we rename Gfir? maybe `Gtirb_ir` or something else

test/cram/gtirb.t
```diff
@@ -0,0 +1,433 @@
...
+       assert false { .opcode = "0x540000c0" };
+       assert boolor(eq(0x40070c:bv64, $PC));
+       goto (%gtirb_3);
+     ];
+     block %gtirb_3 [
+       assume eq(0x40070c:bv64, $PC);
+       assert false { .opcode = "0xd65f03c0" };
+       assert boolor(eq(0x400768:bv64, $PC), eq(0x400704:bv64, $PC),
+        eq(0x4006f8:bv64, $PC));
+       goto (%ret_5,%gtirb_2,%gtirb_1);
```
katrinafyi: this is a `ret` for those curious

lib/gtirb/disas.ml
```diff
@@ -0,0 +1,40 @@
...
+        if i = 0 then Some a else None))
+
+let print_op (opcode : Opcode.t) : string =
+  let opcode_le = Opcode.to_le_bytes opcode in
+  let p_byte (b : char) = Printf.sprintf "0x%02X" (Char.code b) in
+  List.of_seq (String.to_seq opcode_le) |> List.map p_byte |> String.concat " "
+
+let dis_op op =
+  let open Option in
+  let* bin = Lazy.force llvm_mc_bin in
```
katrinafyi: only one use of `let*` doesn't really need the let open

lib/gtirb/gfir_to_bincaml.ml
```diff
@@ -0,0 +1,280 @@
...
+  in
+  let proc, retbl = Procedure.fresh_block ~name:"%ret" ~stmts:[] proc () in
+  let proc =
+    Procedure.map_graph
+      (fun g -> Procedure.G.add_edge g (Procedure.Vert.End retbl) Return)
+      proc
+  in
+  let src = get_vert_block src in
+  let tgt = get_vert_block tgt in
+  let open Gfir.Edge in
```
katrinafyi:
```suggestion
```

lib/gtirb/gfir_to_bincaml.ml
```diff
@@ -0,0 +1,280 @@
...
+  in
+  (* Possible entry addresses *)
+  let requires =
+    entry_addrs
+    |> Iter.map (fun i ->
+        Lang.Expr.BasilExpr.(
+          binexp ~op:`EQ (bvconst (Bitvec.of_int ~size:64 i)) (rvar conf.pc_var)))
+    |> Iter.to_list
+    |> fun conj ->
+    Lang.Expr.BasilExpr.applyintrin ~op:`OR conj |> fun pc_init -> [ pc_init ]
```
katrinafyi, line 258:
```suggestion
    |> Lang.Expr.BasilExpr.applyintrin ~op:`OR
```
then put the whole rhs of `requires = ` inside `[]`.

lib/gtirb/gfir_to_bincaml.ml
```diff
@@ -0,0 +1,280 @@
...
+      Gfir.(
+        verts
+        |> Iter.flat_map (fun v -> G.succ p.cfg v |> List.to_iter)
+        |> Iter.filter_map (function
+          | Vert.Internal uuid -> UUIDMap.find_opt uuid p.code_blocks
+          | Vert.External uuid -> UUIDMap.find_opt uuid all_blocks
+          | Vert.Stmts { uuid } ->
+              Option.bind uuid (fun uuid -> UUIDMap.find_opt uuid all_blocks))
+        |> Iter.map (fun (b : Gtirb.block) -> b.address))
+    with Invalid_argument _ -> Iter.empty
```
katrinafyi, line 273: can we be more precise with this try/catch by only catching the thing which could fail? what exactly could fail here

katrinafyi (approved):
thanks for porting the auxdata decoding.

remind me what was the problem with capstone? i think you mentioned the headers weren't found but the test in the package definition uses headers just fine: https://github.com/Homebrew/homebrew-core/blob/23ed8a378b04d94848042f8bd2f314ffcc8b054b/Formula/c/capstone.rb#L35. homebrew users may just need to set an include path or something. we can fix that up in a follow up  pr.

agle:
> remind me what was the problem with capstone?

I vendored the headers but gave up trying to get it to link to homebrew's install on macos, don't have access to a mac machine. Its still up here https://github.com/agle/bincaml/tree/gtirb-capstone

lib/gtirb/gfir.ml
agle: very nice

katrinafyi:
Merging this will break the Nix until Nixpkgs (and maybe pac-nix) get dune 3.23. It's currently on 3.21 https://search.nixos.org/packages?channel=unstable&query=dune#show=dune.

agle:
Since this seems to be a complete blocker for anyone using nix I've changed the build script to generate and promote the list of diff actions. Its a lot uglier but I guess that's what you've got to do. https://github.com/UQ-PAC/bincaml-test/compare/8a8cd5178397c70e1d62ab046da659ba8370be7e...02b55b556d1d45a028b04bcd26aea0277867d81b#diff-6e68ccebc924f1b47f746c68cc452339d9b277d71c86b2f1ea304c3bb37316af

lib/gtirb/gfir_to_bincaml.ml
agle: its needed for field-based record type disambiguation.

lib/gtirb/disas.ml
agle: I'm replacing it with capstone anyway

lib/gtirb/gtirb.ml
```diff
@@ -0,0 +1,146 @@
...
+  in
+  let rblocks = List.filter_map (chop_block_opcodes ~need_flip) ival_blks in
+  rblocks
+
+let of_file infile =
+  let bytes =
+    let ic = open_in_bin infile in
+    let len = in_channel_length ic in
+    let magic = really_input_string ic 8 in
+    let res = really_input_string ic (len - 8) in
```
katrinafyi: can we use https://ocaml.org/p/containers/3.13/doc/containers/CCIO/index.html#val-read_all_bytes instead? this will let us provide gtirb through pipes as well as files, since pipes don't support in_channel_length

katrinafyi (approved):

PR #161, opened 2026-05-25
Boogie messages

mira-alford:
A small tweak to the boogie backend, has it look out for attributes in ensures/requires/asserts so that boogie error messages can be generated. Also adds messages for the memory encoding invariants.

### DIFF
--- a/lib/backends/boogie.ml
+++ b/lib/backends/boogie.ml
@@ -406,22 +406,24 @@ let pretty_modifies (p : Program.proc) =
     ]
 
 let pretty_expr_attribs e =
+  let open Containers_pp in
   let attrib = Expr.AbstractExpr.get_attrib @@ Expr.BasilExpr.unfix e in
-  pretty_attribute_map ".boogie" attrib
+  if StringMap.mem ".boogie" attrib then
+  sp ^ pretty_attribute_map ".boogie" attrib else text ""
 
 let pretty_ensures (p : Program.proc) =
   let open Containers_pp in
   let spec = Procedure.specification p in
   spec.ensures
   |> List.map (fun s ->
-      text "ensures" ^+ pretty_expr_attribs s ^+ pretty_expr s)
+      text "ensures" ^ pretty_expr_attribs s ^+ pretty_expr s)
 
 let pretty_requires (p : Program.proc) =
   let open Containers_pp in
   let spec = Procedure.specification p in
   spec.requires
   |> List.map (fun s ->
-      text "requires" ^+ pretty_expr_attribs s ^+ pretty_expr s)
+      text "requires" ^ pretty_expr_attribs s ^+ pretty_expr s)
 
 let pretty_procedure_spec (p : Program.proc) =
   let open Containers_pp in
--- a/lib/backends/boogie.ml
+++ b/lib/backends/boogie.ml
@@ -319,7 +319,17 @@ let pretty_statement (s : Program.stmt) =
         ls >|= compose snd pretty_expr |> fill (text "," ^ newline_or_spaces 1)
       in
       nest 2 @@ lhs ^+ text ":=" ^+ rhs
-  | Instr_Assert { body } -> text "assert" ^+ pretty_expr body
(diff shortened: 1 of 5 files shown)


lib/backends/boogie.ml
```diff
@@ -404,17 +405,31 @@ let pretty_modifies (p : Program.proc) =
...
+let pretty_expr_attribs e =
+  match Expr.BasilExpr.unfix e with
+  | RVar { attrib }
+  | Constant { attrib }
+  | UnaryExpr { attrib }
+  | BinaryExpr { attrib }
+  | ApplyIntrin { attrib }
+  | ApplyFun { attrib }
+  | Lambda { attrib }
+  | Let { attrib } ->
```
agle: `AbstractExpr.attrib`

agle (approved):
Neat!

test/cram/malloc_free.t
```diff
@@ -205,20 +217,26 @@
   }
   procedure p$malloc(R0_in: bv64) returns (R0_out: bv64);
     modifies $mem_encoding, $mem, $stack;
-    ensures $me_can_allocate(old($mem_encoding), R0_out, R0_in);
-    ensures ($me_addr_offset($mem_encoding, R0_out) == 0bv64);
-    ensures ($me_alloc_base($mem_encoding, $me_addr_alloc($mem_encoding, R0_out)) == R0_out);
-    ensures ($mem_encoding == $me_allocate(old($mem_encoding), R0_out, R0_in));
+    ensures  $me_can_allocate(old($mem_encoding), R0_out, R0_in);
```
katrinafyi: extra spaces </3

mira-alford: ahhhh yes of course that was going to happen.

test/cram/memory_safety.t
```diff
@@ -20,29 +18,26 @@
...
   Execution trace:
-      out.bpl(332,3): b#inputs
+      out.bpl(367,3): b#inputs
   
   Boogie program verifier finished with 1 verified, 5 errors
 
 
   $ diff after.il after2.il
+  diff: after2.il: No such file or directory
+  [2]
```
katrinafyi: also this is suspicious right? the two files used to be identical and now after2.il doesn't exist?

mira-alford: Yeah, this is a result of the boogie prepass invalidating the IR (a reasonable assumption given no more passes get run). It previously had no issue but now it adds a bunch of attributes which get generated as ir but are not actually part of the grammar and dont get parsed by the parser :)

much bigger problem than I'd like to solve, but given the prepass isn't supposed to generate particularly valid IR i think its probably fine?

mira-alford: Wait actually its the memory encoding adding those invalid attributes for messages... annoying.

mira-alford:
ummm... why is the formatter failing?

test/cram/memory_safety.t
mira-alford: @katrinafyi i can make an issue for this but I think its quite a bit bigger of an issue than this PR was supposed to be for. There's a few things that have attributes but are missing them in the grammar.

katrinafyi: Right, so this led you to delete the `(load-il after.il)  (dump-il after2.il)` lines which leads to this diff error? 

I think anything representable in the AST *should* be valid in the grammar and survive a round-trip, can you file an issue about that or is it something you have your eye on?

agle:
I just pushed a broken container image sorry,  its rebuilding now, I'll rerun the action once I've pushed a new one

test/cram/memory_safety.t
agle: > I think anything representable in the AST should be valid in the grammar and survive a round-trip, can you file an issue about that or is it something you have your eye on?

Seconded,  the attributes aren't present in the grammar for expressions yet.

mira-alford: Done in #162. also removed the leftover diff that shouldn't be there.

PR #163, opened 2026-05-25
Memory allocation intrinsics

### LINKED ISSUE #73: Memory Allocation Intrinsics
1. Introduce an IR instrinsic function to represent zeroing and nonzeroin memory allocation (calloc and malloc) (use `IntrinCall` stmt), probably make the set of intrinsics an ADT or polymorphic variant.
2. Introduce a transform pass that soundly replaces calls to external procedures named `malloc`, `calloc`, `free` with calls to this intrinsic. Be careful of the detail of which registers are preserved or clobbered across the call. Most likely assume an IR with parameter form, and assume the ABI is adhered to.

The goal here is to avoid analyses having to ever match on procedure names and apply a specific semantics; if we do this we need to introduce an intrinsic.

JTrenerry:
Adds memory allocation intrinsics.

I am unsure what to do with free_alloca, should this be called on the exit of a proc when it contains an alloca?

Also unsure what names should be accepted to be transformed.

Also unsure why the attribs are not updating for the calls.

### DIFF
--- a/lib/transforms/intrin_call.ml
+++ b/lib/transforms/intrin_call.ml
@@ -1,0 +1,102 @@
+open Bincaml_util.Common
+open Lang
+open Expr
+
+let transform_proc proc prog =
+  let replace_call (stmt : Program.stmt) : Program.stmt Iter.t =
+    match stmt with
+    | Stmt.Instr_Call { procid; lhs; args; attrib } -> (
+        match ID.name procid with
+        | ( "@malloc" | "@#malloc" | "@calloc" | "@alloca" | "@_malloc"
+          | "@_calloc" | "@_alloca" ) as name ->
+            let modifies =
+              List.filter
+                (fun var -> not @@ String.equal (Var.name var) "R0")
+                (Procedure.specification (Program.proc prog procid))
+                  .modifies_globs
+            in
+            let lhs =
+              StringMap.filter (fun name _ -> String.equal name "R0") lhs
+              |> StringMap.values |> List.of_iter
+            in
+            let args =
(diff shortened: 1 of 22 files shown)


lib/transforms/intrin_call.ml
```diff
@@ -0,0 +1,97 @@
...
+open Expr
+
+let string_to_intrin =
+  Stmt.Intrinsic.(
+    function
+    | "@malloc" | "@#malloc" | "@_malloc" -> Malloc
+    | "@calloc" | "@_calloc" | "@zmalloc" -> Calloc
+    | "@alloca" | "@_alloca" -> AllocStack
+    | "@free" | "@#free" | "@_free" -> Free
+    | _ -> failwith "Unreachable")
```
b-paul, line 12: Could this return an option and have `replace_call` match on the result of this function? that way there isn't duplication of function name strings (though they are also duplicated in lang/stmt.ml so perhaps a better solution exists).

lib/transforms/type_check.ml
```diff
@@ -263,6 +263,49 @@ let type_check stmt_id block_id expr =
...
+    | Malloc | AllocStack | Calloc -> (
+        match (args, params) with
+        | [ size ], [ ptr ] ->
+            let err =
+              if Types.leq (Bitvector 64) (BasilExpr.type_of ptr) then []
+              else
+                [
+                  TypeError
+                    { text = "Wrong type of arg in " ^ to_string intrin_call };
+                ]
```
b-paul, line 282: could you put this if-then-else into a function, since it's quite big and repeated

PR #165, opened 2026-05-25
add locations to stmt attrib sets

agle:


katrinafyi (commented):
what's the idea here? every expression/statement gets a location which is derived from its subexpressions?

agle:
Best effort at the character range encompassing the entire statement, for error reporting relating to a statement, e.g. an assertion fail.

The expressions already had such location tags but its not super precise because we don't have full access to token locations, just what bnfc exposes.

katrinafyi:
Wait so you're talking about character ranges arising from the parsing/tokenising? What are the `%0000035a` labels, then?

agle:
They're BAP's block labels present in the input that were previously not printed ~~(because stmts didn't print their attribs).~~ because the parsed attrib sets were'nt being added to stmts.

katrinafyi (approved):

PR #166, opened 2026-05-25
Represent transform dependencies

### LINKED ISSUE #130: Encode and check transform dependencies

katrinafyi:
This adds an Invariants module for representing transform dependencies on certain named invariants/properties. At the moment, invariants are taken to be whole-program which lines up with the `Proc` transform type which transforms all procedures. 

I need help specifying invariant configs for a lot of the transforms. I've assumed most things just depend on SSA aside from simple structural simplifications which depend on nothing. There is also a problem where most transforms use cram tests which run a single transform on a handmade .il file. This raises invariant warnings because we don't know which properties the test .il file satisfies.

This also only applies to transforms within passes.ml. Things like the Boogie backend aren't in this pass framework and so don't have dependencies. 

Invariants are recorded as an `invariants` attribute (a list of strings) on the Program.t. Program attributes aren't printed into the IL at the moment.

Closes https://github.com/agle/bincaml/issues/130


lib/invariants.ml
```diff
@@ -0,0 +1,116 @@
...
+      *)
+[@@deriving show { with_path = false }, eq, ord]
+
+let read s =
+  match s with
+  | "SSA" -> SSA
+  | "DSA" -> DSA
+  | "Params" -> Params
+  | "ReducibleLoops" -> ReducibleLoops
+  | _ -> failwith (Printf.sprintf "cannot parse string into Invariants.t: %s" s)
```
agle, line 34: ppx_deriving enum can generate this I think

lib/invariants.ml
```diff
@@ -0,0 +1,116 @@
...
+end
+
+type config = {
+  presupposes : t list;
+      (** Invariants which are required to hold {i before} this transform can be
+          applied. *)
+  establishes : t list;
+      (** Invariants which are known to hold this transform. [establishes] and
+          [invalidates] must be disjoint. Invariants which are not present in
+          [establishes] or [invalidates] are assumed to be unchanged. *)
```
agle, line 57: can we just say requires and ensures?

agle (commented):
I think this is fine, the test warnings can be fixed by adding the attribute to the input files in some cases? Do you want me to directly push updated invariant specifications? I was a bit unsure on the invariant living in the attribute at first but its actually a perfect use for it as it very clearly surfaces the reasoning of the pass tool. 

For the boogie etc output passes we can convert them to check passes and have the script run those passes rather than run them directly. They will need to construct a pass that closes over the output file name though. This can be the next PR. We will eventually move towards directly integrating boogie so these are genuinely check passes which just surface alarms.

As the output passes depend on a filename parameter it should be implemented by supporting parametric passes via the script: have function like `create_batch` that takes `sexp -> pass`  like how the script unpacks the sexp to run a command. I have again been thinking it would be nice to write the dsl for for the script parameter format so its easier to add completion support (because last time I wrote a new cmd I couldn't be bothered adding the completion) when we extend it to passes with parameters (e.g. file path completion, procedure/block name completion, pass name completion). And reuse this dsl for the actual parsing, repl hints, completion. I'm not sure how parameters will work with the LSP though---can we get them from the cursor context somehow?

lib/invariants.ml
katrinafyi: I think it only did to and from integer IDs

lib/invariants.ml
katrinafyi: i don't want to overload the terminology

lib/invariants.ml
agle: Ah i was looking at an older library ppx_enum :(

katrinafyi:
Also yeah, feel free to push changes, thanks :)

lib/invariants.ml
agle: IDK if its overloading because it actually means the same thing.

agle (approved):

agle:
I think these conflicts are most easily solved with `git rm -r examples && git commit -m 'rm' && git pull --no-rebase origin main`

katrinafyi:
what is happening here?
```diff
diff --git a/_build/default/test/cram/memory_interproc.t b/_build/default/test/cram/memory_interproc.t.corrected
index f750024..882f546 100644
--- a/_build/default/test/cram/memory_interproc.t
+++ b/_build/default/test/cram/memory_interproc.t.corrected
@@ -11,19 +11,5 @@
   (run-transforms flatten-phis)
   (dump-boogie out.bpl)
   $ boogie out.bpl
-  out.bpl(222,5): Error: a precondition for this call could not be proved
-  out.bpl(101,3): Related location: this is the precondition that could not be proved
-  Execution trace:
-      out.bpl(218,3): b#inputs
-  Memory Error: Invalid Free (not base address)
-  Execution trace:
-      out.bpl(218,3): b#inputs
-  Memory Error: Invalid Free (object not live)
-  Execution trace:
-      out.bpl(218,3): b#inputs
-  out.bpl(230,5): Error: a postcondition could not be proved on this return path
-  out.bpl(112,3): Related location: this is the postcondition that could not be proved
-  Execution trace:
-      out.bpl(218,3): b#inputs
   
-  Boogie program verifier finished with 1 verified, 4 errors
+  Boogie program verifier finished with 2 verified, 0 errors
```

version difference?
```
Boogie program verifier version 3.5.6.0, Copyright (c) 2003-2014, Microsoft.

cvc5 1.3.4
```

PR #167, opened 2026-05-27
Flat memory

mira-alford:
The flat memory encoding has made its way to bincaml after enough procrastination. It even works under the same spec as the split encoding, yay!

Seems to all work well (I'm surprised by how little debugging i had to do actually). The test cases all pass except for one failure to verify in interproc, but that seems like a  whole ordeal and I'm quite happy with half of it verifying. 

I'm a little worried I did something bad to the submodules repo when rebasing accidentally?  but it seems to be working fine. Git is just giving me a scary "submodule exmaples deleted form 8a8cd5" change 💀

### DIFF
--- a/examples/readme.md
+++ b/examples/readme.md
@@ -1,0 +1,1 @@
+examples for github.com/agle/bincaml
--- a/examples/readme.md
+++ b/examples/readme.md
@@ -1,1 +1,0 @@
-examples for github.com/agle/bincaml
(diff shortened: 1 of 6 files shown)


katrinafyi:
has something bad happened to the submodule?

image

mira-alford:
Ok pretty sure I fixed the submodule issue, submodules are scary 😨

agle (approved):

PR #168, opened 2026-05-28
Passes for CHC-based invariant inference

aaronbembenek:
This PR adds passes for inferring program invariants (loop invariants, and pre/postconditions) in the Basil IL using CHC solving. In short, the passes encode the program's IL as CHCs, and then uses a CHC solver to find a model; the invariants are extracted from the CHC model and then added to the IL as assertions.

The two invariant-inference passes are:

- `chc-infer-invariants`: this pass is all or nothing - it finds invariants strong enough to discharge all obligations, or nothing
- `chc-infer-invariants-per-query`: this pass infers invariants on a per-query basis (e.g., per assertion/pre/postcondition), which means that it can find invariants that are strong enough to discharge some, but not all obligations

Right now, we use Z3/Spacer to do the CHC solving; it should be possible to add support for other CHC solvers, like Eldarica, in the future.

The encoding of the IL into CHCs is based on the encoding used in this paper: [Bit-Vector CHC Solving for Binary Analysis and Binary Analysis for Bit-Vector CHC Solving](https://arxiv.org/abs/2603.27107). However, we have augmented the encoding to support pre/postcondition specs for procedures.

- If a procedure has a precondition, a query is generated checking it at all call sites
- If a procedure has a postcondition, a query is generated checking it at the end of that procedure
- The encoding function takes in a predicate `use_specs` over procedures; if `use_specs p` is true for some procedure `p`, then the pre/postconditions for `p` are inlined at call sites of `p`, instead of generating CHCs encoding that control flow goes to the actual body of `p`
- By default, `use_specs p` returns true iff the procedure `p` has non-trivial specifications

~~This PR also adds a `load-store-reduction` pass, which uses some of the pre-processing infrastructure from the Boogie backend to simplify the IL; it is required to run this pass before the CHC invariant inference passes. In particular, we require this sequence of passes to be run first: `full-ssa → load-store-reduction → lambda-lifting`.~~


lib/transforms/chc_infer.ml
```diff
@@ -0,0 +1,823 @@
...
+      BasilExpr.substitute
+        (fun v ->
+          match VarMap.find_opt v in_subst with
+          | Some arg_e -> Some arg_e
+          | None -> VarMap.find_opt v out_subst)
+        e
+    in
+    List.iter
+      (fun ens ->
+        Encoder.add_premise enc (encode_expr enc (apply_full_subst ens)))
```
agle: List.iter (apply_full_subts %> encode_expr %> Encoder.add_premise) spec.ensures

lib/transforms/chc_infer.ml
```diff
@@ -0,0 +1,823 @@
...
+    before doing the Δ lookup.
+
+    [use_spec] decides per-callee how the call site connects to the callee: when
+    true, the caller assumes the callee's [ensures] post-call and does not refer
+    to the callee's [enter]/[exit] predicates; when false, the caller emits a
+    [C ⟹ enter⟨f⟩(args)] clause and uses [exit⟨f⟩] as a premise. Spec checking
+    (precondition queries at the call site, postcondition queries on the body)
+    is performed independently of [use_spec] whenever the callee has a
+    [requires] or [ensures]. *)
+let encode_block ~(use_spec : Program.proc -> bool) (prog : Program.t)
```
agle: Most of the inner functions in this function should be made top-level so its clear they only depend on their in-parameters rather than any prior local definitions. In particular translation of regular statements, calls, and blocks should be different top-level functions.

lib/transforms/chc_infer.ml
```diff
@@ -0,0 +1,823 @@
...
+  List.iter (fun pr -> p (declare_predicate pr)) preds;
+  List.iter (fun c -> p (clause_to_sexp c)) clauses;
+  p (list [ atom "check-sat" ])
+
+let dump_to_file (prog : Program.t) (path : string) : unit =
+  Logs.info (fun m -> m "Dumping CHC clauses to %s" path);
+  let preds, clauses = encode_program prog in
+  CCIO.with_out path (fun oc -> emit_chc_problem oc preds clauses)
+
+type solve_result = Sat | Unsat | Unknown
```
agle: From here below could be a separate file devoted to the solver interaction.

test/cram/load_store_reduction.t
```diff
@@ -0,0 +1,28 @@
...
+  $ grep -c -E '(store|load) le ' load_store_reduction.il
+  2
+
+After the pipeline, no addressed store/load instructions remain:
+
+  $ grep -E '(store|load) le ' load_store_reduction_out.il || echo "no load/store instructions"
+  no load/store instructions
+
+The reduction introduces uninterpreted `load64_le`/`store64_le` function
+declarations:
```
agle: They are functions with a definition, not uninterpreted. They seem to not be defined/declared in the SMT/HORN solver so the loop inference crashes if you put `(run-transforms chc-infer-invariants)` at the end of this sexp file. I'm not sure if that's intended to be in scope for this PR but it seems like the intention for this pass.

lib/transforms/chc_infer.ml
```diff
@@ -0,0 +1,823 @@
...
+  | None -> { blocks = []; returns = false }
+  | Some g ->
+      Procedure.G.fold_succ
+        (fun v acc ->
+          match v with
+          | Procedure.Vert.Begin id -> { acc with blocks = id :: acc.blocks }
+          | Return | Exit -> { acc with returns = true }
+          | _ -> failwith "chc_infer: bad CFG structure")
+        g (Procedure.Vert.End block_id)
+        { blocks = []; returns = false }
```
agle: `Procedure.blocks_succ proc %> Iter.map fst %> Iter.to_list` and 

```ocaml
let block_returns p i =
  Option.to_iter (Procedure.graph p)
  |> Iter.exists (fun graph ->
      Iter.from_iter (fun f -> G.iter_succ f graph (End i))
      |> Iter.exists (function Vert.Return -> true | _ -> false))

```

Exit is intended to represent the early termination or abort path so doesn't establish the post condition, it probably shouldn't be included.

lib/transforms/chc_infer.ml
```diff
@@ -0,0 +1,823 @@
+(** CHC encoding of the IR.
+
+    This module encodes a program as a system of constrained Horn clauses (CHCs)
+    over quantifier-free bit vectors. See {!Lang.Expr_smt.SMTLib2} for
+    expression-level encoding and {!Bincaml_util.Smt} for the SMT-LIB sexp
+    helpers used below. *)
```
agle: Could link the preprint in the comment as well as the PR.

lib/transforms/chc_infer.ml
```diff
@@ -0,0 +1,823 @@
...
+
+type clause = {
+  vars : Var.t list;
+  premises : Sexp.t list;
+  head : Sexp.t option;
+}
+(** A constrained Horn clause.
+    - [vars] are the universally-quantified binders for the [forall];
+    - [premises] are the conjuncts of the antecedent;
+    - [head] is [Some sexp] for a rule and [None] for a query (head = false). *)
```
agle: Field doc comments should be attached to the field.

```ocaml 
type clause = {
  vars : Var.t list; (** universally-quantified binders for the [forall] *)
  premises : Sexp.t list; (** conjuncts of the antecedent *)
  head : Sexp.t option; (** [Some sexp] for a rule and [None] for a query (head = false) *)
}
```

agle (changes requested):
The code generally looks fine but the load/store encoding seems to not work yet. For the final version it would be good to achieve parity with the python implementation up to the loopy benchmarks if its possible. We can set the test up like here https://github.com/UQ-PAC/bincaml-test/tree/main/gtirb, happy to help with that if needed.

`full-ssa → load-store-reduction → lambda-lifting` this sequence requirement can also be expressed now with the new pass dependency system.

lib/transforms/chc_infer.ml
aaronbembenek: I'll move the CHC solver interaction stuff that is Basil-agnostic to a new file.

test/cram/load_store_reduction.t
aaronbembenek: You're totally right - the CHC encoder was not handling memory operations correctly. I've fixed that.

aaronbembenek:
> `full-ssa → load-store-reduction → lambda-lifting` this sequence requirement can also be expressed now with the new pass dependency system.

This has been simplified as load-store reduction and lambda-lifting now happen internally in the CHC encoding process. I have added SSA as a dependency. Technically, the CHC pass works fine without SSA. However, I think it still makes sense to include it as a dependency, so that people run SSA before it. If SSA is run after it, the loop invariants can get messed up: the assertions (encoding the invariants) are no longer the first statement in the block, which seems to have the effect of Boogie no longer accepting them.

aaronbembenek:
> For the final version it would be good to achieve parity with the python implementation up to the loopy benchmarks if its possible. We can set the test up like here https://github.com/UQ-PAC/bincaml-test/tree/main/gtirb, happy to help with that if needed.

Loopy tests added here: https://github.com/UQ-PAC/bincaml-test/pull/1

The CHC pass completes on 460/468 benchmarks. For 3 benchmarks, Z3 crashes; for the other 5, bincaml SSA crashes (I can open an issue).

CHC solver results (timeout of 10s):

- 327 sat
- 23 unsat
- 110 unknown

Boogie verifies 316 benchmarks (annotated with the inferred invariants).

lib/transforms/chc_infer.ml
```diff
@@ -0,0 +1,879 @@
...
+                          let m =
+                            VarMap.of_list (List.combine bound_vars args)
+                          in
+                          BasilExpr.substitute
+                            (fun v -> VarMap.find_opt v m)
+                            in_body
+                      | _ -> BasilExpr.fix node)
+                  | None -> BasilExpr.fix node)
+              | _ -> BasilExpr.fix node)
+          | _ -> BasilExpr.fix node)
```
agle, line 592: This should be much simpler, with BasilExpr.unfix3 and BasilExpr.rewrite_down rather than cata; it should have one match on multiple levels of the ast rather than 4 levels of nested match. It probably makes mores sense using downard recursion rather than upward.

lib/lang/expr_smt.ml
```diff
@@ -192,12 +235,54 @@ module SMTLib2 = struct
...
+        in
+        let fresh_var typ =
+          let rec go i =
+            let n = Printf.sprintf "x_%d" i in
+            if StringSet.mem n !used then go (i + 1)
+            else (
+              used := StringSet.add n !used;
+              Var.create n typ)
+          in
+          go 0
```
agle: This is messy, for each recursive call it will have to iterate through all the previously allocated indexes to check they are in the map. It would be sensible to use ID.generator. Call `decl_or_get` to reserve the existing names, and `fresh` to return a new name (ID.name to get the unique string for the id). There is an ID.generator stored in the procedure for allocating local variables, it could use this when appropriate.

lib/lang/expr_smt.ml
```diff
@@ -181,6 +210,20 @@ module SMTLib2 = struct
     | `Atom "false" -> Some (BasilExpr.boolconst false)
     | `Atom e when Int.of_string e |> Option.is_some ->
         Some (BasilExpr.intconst (Z.of_string e))
+    | `Atom s
+      when String.length s > 2 && Char.equal s.[0] '#' && Char.equal s.[1] 'b'
+      ->
+        let digs = String.sub s 2 (String.length s - 2) in
+        let size = String.length digs in
+        let v = Z.of_string ("0b" ^ digs) in
+        Some (BasilExpr.bvconst (Bitvec.create ~size v))
```
agle: Would it be better to disable prettyprinting of bitvectors in the solver to parse a simpler representation?

lib/transforms/chc_infer.ml
```diff
@@ -0,0 +1,879 @@
...
+(** Encode a single block. Walks the block's regular statements (threading the
+    encoder state), then appends one rule clause per outgoing edge (block
+    successor or procedure return). Phi nodes in a successor are resolved by
+    substituting each phi target with the appropriate rhs variable for this edge
+    before doing the Δ lookup.
+
+    [use_spec] decides per-callee how a call site connects to the callee: when
+    true, the caller assumes the callee's [ensures] post-call and does not refer
+    to the callee's [enter]/[exit] predicates; when false, the caller emits a
+    [C ⟹ enter⟨f⟩(args)] clause and uses [exit⟨f⟩] as a premise. Spec checking
```
agle: I'm a bit suspect of the kind of unneccessary use of unicode math characters, they'll not display well on a lot of terminal text editors.

lib/transforms/chc_solve.ml
```diff
@@ -0,0 +1,121 @@
...
+          | `List [ `Atom "define-fun"; `Atom name; `List params; _ret; body ]
+            ->
+              let params =
+                List.filter_map
+                  (fun p ->
+                    match p with
+                    | `List [ `Atom v; sort ] -> Some (v, sort)
+                    | _ -> None)
+                  params
+              in
```
agle: Is this missing the support for quoted idents? Also lots of cases are dropped, if we're confident there is an exact form it might be good to fail so its easier to debug when we run into issues with a different solver, etc.

lib/transforms/chc_solve.ml
```diff
@@ -0,0 +1,121 @@
...
+  let binders = List.map (fun v -> list [ var_atom v; var_sort v ]) c.vars in
+  let premise = SmtExpr.bool_ands c.premises in
+  let body =
+    match c.head with
+    | Some h -> SmtExpr.bool_implies premise h
+    | None -> SmtExpr.bool_not premise
+  in
+  list [ atom "assert"; forall_ binders body ]
+
+type solve_result = Sat | Unsat | Unknown
```
agle: Could put the model in the SAT case rather than returning a tuple

lib/transforms/chc_solve.ml
```diff
@@ -0,0 +1,121 @@
...
+  let s = Solver.create Bincaml_util.Smt.Config.z3 in
+  Solver.set_option s ":fp.xform.slice" "false";
+  Solver.set_option s ":fp.xform.inline_eager" "false";
+  Option.iter
+    (fun t -> Solver.set_option s ":timeout" (string_of_int t))
+    timeout_ms;
+  Solver.set_logic s "HORN";
+  s
+
+type model_def = string * (string * Sexp.t) list * Sexp.t
```
agle: Could have named fields since these types don't convey any information.

lib/transforms/chc_solve.ml
aaronbembenek: Assuming that quoted idents are parsed as an atom by the Sexp parser, this should handle them okay. I'm not sure that is the case in general (in fact, I'd be surprised if it is), but it should be fine in practice, since presumably the parameter names follow pretty standard syntax, aren't quoted, and thus don't include tricky characters like " " or ")" that might throw off the S-expression parser.

I've changed this code so that it raises an exception if models/function definitions do not have the expected shape.

lib/transforms/chc_infer.ml
aaronbembenek: I've removed all unicode from comments.

lib/lang/expr_smt.ml
aaronbembenek: ID.generator does make this a lot cleaner! I've just created a local generator with ID.make_gen so that expr_of_smt does not need to be called with a procedure's ID.generator. It still guarantees that no Basil variables are shadowed by the name given variables bound at quantifiers.

lib/lang/expr_smt.ml
aaronbembenek: I'd be in favor in leaving this as is; it's not that much extra code, and it gives us more flexibility if we end up using solvers that do not have the option to turn off this syntax (which is part of the SMT-LIB standard).

agle (approved):

PR #170, opened 2026-06-08
Gtirb capstone

### LINKED ISSUE #186: Linux CI job doesn't run inline tests?
You can checkout main and add this buggy code to `test/transforms/test_aslp.ml`
```ocaml
let%expect_test "a" =
  print_endline "a"
```
Then, if you run 
```
dune runtest --force --profile=release test/transforms/  
```
, like the linux CI does, this will not show any errors.

The problem seems to be when `--profile=release` is included but idk why. Executable tests still seem to run. Removing the release mode argument shows the test error.

This leads to confusing differences between the macos and linux CI, since they use slightly different commands and the macos one doesn't use release mode.

agle:
Adds capstone depext to use C library to disas opcodes.  (Substantially faster than subprocessing to LLVM- we can now actually disas all the tests in reasonable time).


capstone_arm64_disas.opam.template
```diff
@@ -0,0 +1,9 @@
+depexts: [
+  ["capstone-devel"] {os-family = "fedora"}
+  ["capstone-dev"] {os-distribution = "alpine"}
+  ["libcapstone-dev"] {os-family = "debian" | os-family = "ubuntu"}
+  ["capstone"] {os = "macos" & os-distribution = "macports"}
+  ["capstone"] {os = "macos" & os-distribution = "homebrew"}
+  ["capstone"] {os = "freebsd"}
+  ["capstone"] {os-distribution = "nixos"}
+]
```
katrinafyi, line 9: can we put the capstone bindings into a separate dune package (even if it stays inside this repo).?

lib/gtirb/cap/capst.c
```diff
@@ -0,0 +1,53 @@
+
+
```
katrinafyi:
```suggestion
```

capstone_arm64_disas/capst.c
```diff
@@ -0,0 +1,53 @@
...
+
+#include <inttypes.h>
+#include <stdio.h>
+
+#include <caml/alloc.h>
+#include <caml/custom.h>
+#include <caml/memory.h>
+#include <caml/mlvalues.h>
+
+#include <capstone/capstone.h>
```
katrinafyi, line 10: is this some exapmle code copied from somewhere? if so, can we add a link to the source for reference?

capstone_arm64_disas/dune
```diff
@@ -0,0 +1,27 @@
...
+
+(library
+ (public_name bincaml.gtirb.capstone)
+ (name capstone)
+ (foreign_stubs
+  (language c)
+  (include_dirs
+   (include includes.gen))
+  (names capst))
+ (c_library_flags -lcapstone -I/opt/homebrew/include))
```
katrinafyi: what's happening here? is there a difference between `-I` and `include_dirs`? why is one conditionally computed and the other not?

capstone_arm64_disas/dune
```diff
@@ -0,0 +1,27 @@
+(rule
+ (target includes.gen)
+ (enabled_if
+  (= %{system} "macosx"))
+ (action
+  (with-stdout-to
+   includes.gen
+   (echo "/opt/homebrew/include"))))
```
katrinafyi: `$(brew --prefix)/include` is probably more robust. homebrew does technically support installing to different paths, though it's very rarely used.

capstone_arm64_disas.opam.template
agle: That doesn't avoid it being a dependency

katrinafyi: .. yeah? I just think the depexts directly in the bincaml opam package are kinda messy.

capstone_arm64_disas/dune
agle: -I is unneccessary I was just hoping to use it to avoid the conditional stuff at one point

capstone_arm64_disas/dune
agle: monkey paw curls https://github.com/agle/bincaml/pull/170/changes/65031163811852d8a4af028a08de40b888e491e4

capstone_arm64_disas/capst.c
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  cs_close(&handle);
+
+  return 0;
+}
+
+CAMLprim value disas_arm64_op(value v) {
+  CAMLparam1(v);
+  const uint8_t *op = Bytes_val(v);
```
katrinafyi, line 55: this is unsafe if given a Bytes with less than 4 bytes. ik it's private but comment or guard against.

katrinafyi (approved):

PR #171, opened 2026-06-09
Fix some broken passes

b-paul:
the simplify batch transform had multiple ssa requirements so this fixes that.

the linear const pass had an incorrect inverse computation, and hence an incorrect join leading to an infinite loop

interprocedural dead store elimination is now annotated as needing ssa (it uses the ssa ide solver)

### DIFF
--- a/lib/analysis/linear_const.ml
+++ b/lib/analysis/linear_const.ml
@@ -73,9 +73,9 @@ module LF = struct
       inverses [mod 2^n].
 
       This computes the join of [ax + b] and [cx + d]. According to the paper we
-      can compute this as [(ax + b) join (a * l + b)], where [l = (b-d)/(c-a)], and
-      we can compute this only when [c-a] has an inverse [mod 2^n], i.e. only
-      when [c-a] is odd (as hence it is coprime to 2^n). *)
+      can compute this as [(ax + b) join (a * l + b)], where [l = (b-d)/(c-a)],
+      and we can compute this only when [c-a] has an inverse [mod 2^n], i.e.
+      only when [c-a] is odd (as hence it is coprime to 2^n). *)
   let compute_join a b c d =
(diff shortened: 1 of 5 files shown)


lib/analysis/linear_const.ml
```diff
@@ -69,13 +69,15 @@ module LF = struct
...
   let compute_join a b c d =
     let bd = Bitvec.value (Bitvec.sub b d) in
-    let g, s, t = Z.gcdext (Bitvec.value (Bitvec.sub c a)) bd in
-    if Z.divisible g bd then
-      let l0 = Bitvec.create ~size:(Bitvec.size a) (Z.mul s bd) in
+    let ca = Bitvec.value (Bitvec.sub c a) in
+    let pow = Z.pow (Z.of_int 2) (Bitvec.size a) in
+    if Z.congruent ca Z.one pow then
+      let cainv = Z.invert ca pow in
+      let l0 = Bitvec.create ~size:(Bitvec.size a) (Z.mul cainv bd) in
```
katrinafyi: Hmm what's this meant to do? 

If `ca = 1 (mod pow)` then isn't the inverse of ca mod pow also just 1 or am I crazy?

b-paul: OH that was meant to be mod 2 ... I guess this is still sound since this branch basically never gets hit (the inverse of ca exists mod 2^n iff the number is odd)

katrinafyi: Cool cool. I've pushed a change to make it a doc comment and try to clarify the join precedence. Let me know if that's okay.

katrinafyi (approved):

lib/analysis/linear_const.ml
b-paul: thank you! that's probably a better idea than hanging comments

PR #176, opened 2026-06-16
nix: fix after gtirb loader

### LINKED ISSUE #175: Nix protobuf
Flake is missing `protoc`

katrinafyi:
Closes https://github.com/agle/bincaml/issues/175

We get "Error: Unbound value Bool.to_string" in Nix but not outside of Nix and this seems to be fixed by using `Stdlib.Bool` so that's what we're gonna use.

I also disable tests because the `  src = nix-gitignore.gitignoreSource [ "nix" "flake.nix" "flake.lock" ] ./..;` seems to exclude the submodule for whatever reason. Tests are still enabled in the dev shell for nixos users.

### DIFF
--- a/lib/gtirb/gfir.ml
+++ b/lib/gtirb/gfir.ml
@@ -40,8 +40,8 @@ module Edge = struct
   let to_attrib { conditional; direct; type' } =
     StringMap.of_list
       [
-        (".conditional", `String (Bool.to_string conditional));
-        (".direct", `String (Bool.to_string direct));
+        (".conditional", `String (Stdlib.Bool.to_string conditional));
+        (".direct", `String (Stdlib.Bool.to_string direct));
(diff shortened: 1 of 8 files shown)


lib/gtirb/gfir.ml
```diff
@@ -40,8 +40,8 @@ module Edge = struct
   let to_attrib { conditional; direct; type' } =
     StringMap.of_list
       [
-        (".conditional", `String (Bool.to_string conditional));
-        (".direct", `String (Bool.to_string direct));
+        (".conditional", `String (Stdlib.Bool.to_string conditional));
+        (".direct", `String (Stdlib.Bool.to_string direct));
```
agle, line 44: This is ridiculous.

agle (approved):

PR #177, opened 2026-06-16
Data structure analysis

b-paul:
This PR implements data structure analysis!

Not included in this pr:
- Indirect call resolution
- Properly updated node flags (which store the information further transforms/analyses would actually want to read)
- Any transforms using the analysis results

Included is a formal model of a dsgraph as described in the recent dsa paper, and an efficient union find based representation (and this unfortunately means the code is imperative and full of reference stuff...). The two representations were compared with a QCheck_Stm approach, until a terrifying edge case showed equality checking the two models was infeasible with the present approach (though prior to said edge case occurring, many bugs were caught!) The proptests still exist as a semi-exhaustive check of invariants that the efficient representation imposes.


test/analysis/test_dsa.ml
```diff
@@ -0,0 +1,441 @@
...
+          Printf.sprintf "MakeNode (%d, %s)" n
+            (List.to_string ~start:"[" ~stop:"]" ~sep:"; " Interval.dbg is)
+      | CountCells n -> Printf.sprintf "CountCells %d" n
+      | CellWidths n -> Printf.sprintf "CellWidths %d" n
+      | MakeEdge (n, a, b) -> Printf.sprintf "MakeEdge (%d, %d, %d)" n a b
+      | Join (n, a, b) -> Printf.sprintf "Join (%d, %d, %d)" n a b
+      | UnifyAll n -> Printf.sprintf "UnifyAll %d" n
+      | Copy (f, t, cs) ->
+          Printf.sprintf "Copy (%d, %d, [%s])" f t
+            (List.to_string ~sep:"; " Int.to_string cs)
```
agle: `[@@deriving show]`should do this

lib/passes.ml
```diff
@@ -465,6 +467,12 @@ module PassManager = struct
         doc = "Replace gamma expressions with gamma variables";
         invariants = Invariants.needs ~invalidates:[ SSA ] [];
       };
+      {
+        name = "data-structure-analysis";
+        apply = Prog Analysis.Dsa.dsa_dots;
+        doc = "";
+        invariants = Invariants.needs ~invalidates:[ SSA ] [];
```
agle: This seems backwards, it needs SSA and invalidates nothing.

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+  let ( != ) a b = not @@ (a = b)
+  let get_flag idx f = (f >> idx & 1) != 0
+  let set_flag idx f = f || 1 << idx
+  let clear_flag idx f = f & Int.lognot (1 << idx)
+
+  (* CCBitField didn't have a join function :( *)
+  let join f f' = f || f'
+
+  (* Should be used like `get_flag heap f` or `clear_flag unknown f` *)
+  let heap = 0
```
agle: `let[@inline]`

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+  type t = int
+
+  let empty = 0
+  let ( >> ) = Int.shift_right_logical
+  let ( << ) = Int.shift_left
+  let ( || ) = Int.logor
+  let ( & ) = Int.logand
+  let ( != ) a b = not @@ (a = b)
+  let get_flag idx f = (f >> idx & 1) != 0
+  let set_flag idx f = f || 1 << idx
```
agle: `let[@inline]`

Could also hide the idx type with an interface so it can only take the result one of the label name functions.

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+  let shift off = function
+    | Bot -> Bot
+    | Top -> Top
+    | Interval (a, b) -> Interval (a + off, b + off)
+
+  let width = function
+    | Bot -> None
+    | Top -> None
+    | Interval (a, b) -> Some (Z.to_int (b - a + one))
+end
```
agle: This is a generic domain we should factor it into a separate file.

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+                    Expr.BasilExpr.rvar @@ StringMap.find s fout ))
+            in
+            let args =
+              StringMap.to_list args
+              |> List.map (fun (s, actual) ->
+                  (actual, Expr.BasilExpr.rvar @@ StringMap.find s fin))
+            in
+            Some (Call { lhs; args; callee_id = procid })
+        | _ -> None)
+end
```
agle: This is nice and neat

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+module Constraint = struct
+  type 'e t =
+    | Mem of { addr : 'e; value : 'e; size : int }
+    (* Actual * Formal pairs *)
+    | Call of { lhs : ('e * 'e) list; args : ('e * 'e) list; callee_id : ID.t }
+  [@@deriving eq]
+
+  (** Special funky map where the second function evaluates within a procedure
+      id for call constraint callee formal arguments *)
+  let map f g c =
```
agle: Could be more clearly communicated  by type annotation / function and argument names

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+  let show f =
+    (if get_flag heap f then "H" else "")
+    ^ (if get_flag stack f then "S" else "")
+    ^ (if get_flag global f then "G" else "")
+    ^ (if get_flag unknown f then "U" else "")
+    ^ (if get_flag modified f then "M" else "")
+    ^ (if get_flag read f then "R" else "")
+    ^ (if get_flag complete f then "C" else "")
+    ^ if get_flag collapsed f then "O" else ""
+end
```
agle: These are directly lifted from the paper- can we document the extent to which we use them and what they mean in our (binary) context?

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+
+(* TODO decide on how flags should be tested (if at all) maybe only test in expect tests
+          and also symbolic bases probably*)
+(* TODO put this in a separate file somehow *)
+
+(** A purely functional (and less efficient) representation of DSGraphs to be
+    used as a model for testing against, which is based on the DSA paper! It's
+    also very ugly because it is based on lots of giant set comprehensions but
+    oh well... *)
+module FormalDSGraph = struct
```
agle: New file, can make subdir for DSA analysis and use the dune `(include_subdirs qualified)` to treat each file as a module as appropriate; https://dune.readthedocs.io/en/stable/reference/dune/include_subdirs.html

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+  (** Get the union find parent of this cell *)
+  let rec find (c : cell) =
+    match !c with
+    | Path c' ->
+        let p = find c' in
+        c := Path p;
+        p
+    | Cell _ -> c
+
+  (** Get the union find parent of this cell *)
```
agle: This node

lib/analysis/dsa.ml
agle: Make parametric in a generic `Domain` interface defined as the current interface of Interval---it only needs a minimal set of operations on the value domain and doesn't inherently require interval.

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+
+  let shift off = function
+    | Bot -> Bot
+    | Top -> Top
+    | Interval (a, b) -> Interval (a + off, b + off)
+
+  let width = function
+    | Bot -> None
+    | Top -> None
+    | Interval (a, b) -> Some (Z.to_int (b - a + one))
```
agle: Dead code

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+        let cells = aux r.cells in
+        match cells with
+        | [] -> None
+        | [ c ] -> Some c
+        | c :: cs ->
+            List.iter (join c) cs;
+            Some c)
+
+  (** Get all cells corresponding to a value set. *)
+  let cells_of (v : Sva.SymAddrSetLattice.t) (g : t) : cell list =
```
agle: Take an iter of value domain values rather than a set lattice- it doesn't look like the lattice ops on the set are used. 

```
    snd @@ Sva.SymAddrSetLattice.to_list v
    |> List.filter_map (fun (sb, i) ->
        let i = Interval.of_wint i in
```

Should all live outside the DSGraph so its not so tightly coupled to the sva analysis type.

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+  (** A graph, which keeps track of all nodes and also which nodes come from
+      each symbolic base (there is duplicate information). *)
+
+  let empty_graph () = { nodes = []; node_map = SBMap.empty }
+  let nodes (g : t) = g.nodes
+  let node_map (g : t) = g.node_map
+  let id_gen = ID.make_gen ()
+
+  (** Get the union find parent of this cell *)
+  let rec find (c : cell) =
```
agle: `find_cell`

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+    match !c with
+    | Path _ -> shift off (find c)
+    | Cell r -> r.offsets <- Interval.shift off r.offsets
+
+  (** Set the node of this cell's parent *)
+  let rec set_node node (c : cell) =
+    match !c with Path _ -> set_node node (find c) | Cell r -> r.node <- node
+
+  (** Set the pointees of this cell's parent *)
+  let rec set_pointees pointees (c : cell) =
```
agle: As far as I can tell this is only going to recurse by one, and the actual iteration is in the call to `find`--- which is an identity for the root node (Cell); could we make this function non recursive and call find before the function that sets the pointees? I.e. 

```ocaml
    find c |> function (Cell f) -> f.pointees <- pointees
```

Its a bit scary seeing rec everywhere and wondering if it will terminate but its really just traversing one level of the union-find tree/forest

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+
+  (** Get the id of this node's parent *)
+  let rec id (n : node) =
+    match !n with NodePath _ -> id (fst @@ find_node n) | Node r -> r.id
+
+  (** Set the cells of this node *)
+  let rec set_cells cells (n : node) =
+    match !n with
+    | NodePath _ -> set_cells cells (fst @@ find_node n)
+    | Node r -> r.cells <- cells
```
agle: This is similar, to prev comment, I'm not sure this pattern is neccessary as find does the same check for the root variant. Leave the union find `find` function (or a helper) responsible for unwrapping the nodes (incl path resolution) and encode it in the type system by returning the argument to the leaf variant constructor, then these functions are simply `(find_node n).cells <- cells` with no recursion.

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+    | Node r -> r.flags <- NodeFlags.join r.flags flags
+
+  (** Make the second cell point to the first *)
+  let rec join_paths (c1 : cell) (c2 : cell) =
+    match (!c1, !c2) with
+    | Path _, Path _ -> join_paths (find c1) (find c2)
+    | Path _, _ -> join_paths (find c1) c2
+    | _, Path _ -> join_paths c1 (find c2)
+    | Cell r, Cell { pointees = p } ->
+        if not @@ CCEqual.physical c1 c2 then c2 := Path c1
```
agle: Similarly could probably call find first and only handle the Cell, Cell case.

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+          | Some n ->
+              n1 := Node { r1 with cells = n; flags };
+              n2 := NodePath (n1, off);
+              check_valid_node n1;
+              check_valid_node n2
+          | _ -> ())
+
+  (** Inserts the cell into the node. *)
+  let rec insert node cell =
+    match !node with
```
agle: Make nonrec and call find on node and cell first.

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+  let nodes (g : t) = g.nodes
+  let node_map (g : t) = g.node_map
+  let id_gen = ID.make_gen ()
+
+  (** Get the union find parent of this cell *)
+  let rec find (c : cell) =
+    match !c with
+    | Path c' ->
+        let p = find c' in
+        c := Path p;
```
agle: I think it could probably directly assign `c := p`? Its still a ref so should point to the same physical copy of the root. Worth a comment if there's a reason the intermediate path is necessary.

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+    | [ x ] -> Some x
+    | x :: xs ->
+        List.iter (join x) xs;
+        Some x
+    | [] -> None
+
+  (** Create a copy of the given node with all reachable nodes and cells from
+      pointees copied recursively. The copy of the given node is returned.
+
+      `sbs` will update the graph's node_map with the newly copied node assigned
```
agle: [sbs]

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+      said symbolic bases. It is assumed that no node with any given sb exists
```
agle: `sb` \in `sbs`?

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+  (** Get the cells of this node's parent *)
+  let rec cells (n : node) =
+    match !n with NodePath _ -> cells (fst @@ find_node n) | Node r -> r.cells
+
+  (** Get the flags of this node's parent *)
+  let rec flags (n : node) =
+    match !n with NodePath _ -> flags (fst @@ find_node n) | Node r -> r.flags
+
+  (** Get the id of this node's parent *)
+  let rec id (n : node) =
```
agle: Needs a more descriptive name; confusing with `Fun.id`

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+    (* Join the cells belonging to the same node, and then join each node
+       together. This avoids imprecision from joining cells from other nodes,
+        and residual cells around the rest of the node getting merged. *)
+    let join_pointees ps =
+      let node_cells =
+        List.fold_left
+          (fun acc c ->
+            let id = id (node_of c) in
+            let cur = IDMap.get_or id acc ~default:[] in
+            IDMap.add id (c :: cur) acc)
```
agle: IDMap.update

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+            join c c';
+            Some c')
+      None cells_copy
+end
+
+(** Construct the local-phase graph of the given procedure. It will be ensured
+    that all loaded/stored addresses, and all formal in/out params are unified,
+    however actual params of calls will not be unified for precision and should
+    be unified at use in the BU/TD phases. *)
+let make_local_graph proc sva
```
agle: This is quite neat; it would be cool to put all the SVA-related DSA analysis construction stuff in its own file.

lib/analysis/dsa.ml
```diff
@@ -0,0 +1,1481 @@
...
+  g
+
+(** Get all callees called by the given procedure *)
+let callees p =
+  Procedure.blocks_to_list p |> List.to_iter |> Iter.map snd
+  |> Iter.flat_map Block.stmts_iter
+  |> Iter.filter_map (function
+    | Stmt.Instr_Call { procid } -> Some procid
+    | _ -> None)
+  |> IDSet.of_iter |> IDSet.to_iter
```
agle: Iter.uniq should be more efficient if you pass it the right hash/ord function. Could also use Procedure.CallGraph but maybe this is simpler.

agle (commented):
Hi, I haven't gotten all the way through yet so I will spend some more time later to look at it (particulary the solvers). Theres a few nit picks but the main thing is it would be nice to break into separate files and have the DSA module be clearly independent of the value-set/data flow analysis domain and the SVA domain. It would be good to take a module parameter for the interval---If hypothetically we use a different domain in place of Interval I don't know if it would realistically be more precise since the operations on it are so limited, but if its broken up it looks like the refactor to make it a module parameter looks not too hard.

lib/analysis/dsa.ml
b-paul: do you know if there's a way to make a module defined from a file to be a functor? (i have moved the formaldsgraph to its own file)

agle: Hm no I think you'll have to define a functor in the file. E.g. in the standard library you have `map.ml` with a `Make` functor.

agle: We discussed this and decided not to

agle (approved):

PR #178, opened 2026-06-16
Initial skeleton of Aslp lifter transform

katrinafyi:
I want to write the Aslp lifter pass as an ordinary transform that operates on the opcode-annotated AST and returns a new AST. 

To fulfill the imperative IBI, we use a first class module with a `ref` parameter to the mutable state. The mutable state is a really restricted "block" graph that aims to represent both within-instruction detail and across-instruction detail, but all of this is within a particular code block. Control flow across code blocks will be handled separately (next).

For the PC correspondence, I want to use the offline lifter in PC mode which means that all PC assignments produced by ASLp will be resolved to concrete addresses. Then, it is a simple matter of tracking conditions leading up to each PC assignment within the semantics, then attaching this as a guard in front of successor code blocks.

This is just an initial PR for feedback on the design. I'm also iffy about putting it in one big `aslp.ml` file in transforms, would there be a better place to put this, especially since it looks like we'll have multiple big modules?


katrinafyi:
the ci needs the PAC opam repository. should we add to the ci commands or is there a better way to specify it in the opam file somewhere?

agle:
> I want to write the Aslp lifter pass as an ordinary transform that operates on the opcode-annotated AST and returns a new AST.

I guess we should generate the initial IR encoding of each opcode as `call Eval(opcode);` rather than assert/assume false with attributes. This makes the correctness property very obvious-- assuming (aslp(asli, MRA)(op) = asli(op)) you get the bi-simulation `eval(op) ~ eval(lift(op))` per-block/opcode directly. Wheras a program that is only magic cannot refine anything (specifically the lifted program). Although in practice/code-wise I guess this is just an aesthetic change.

agle (approved):

PR #179, opened 2026-06-17
Fixed width zarith bvops

agle:
Factor out fixed width arith encoding from the bitvector representation.

### DIFF
--- a/lib/util/bitvec.ml
+++ b/lib/util/bitvec.ml
@@ -1,5 +1,4 @@
 open Containers
-module I = Fixed_width_arith
 
 (* workaround: ZArith library doesn't like zero-length extracts *)
 let checked_extract f v off len = if len > 0 then f v off len else Z.zero
@@ -81,7 +80,7 @@ let of_string i =
 let size_is_equal a b = assert (size a = size b) [@@inline always]
 
 let size_if_equal a b =
-  if size_is_equal a b then size a else failwith "bv binop widths differ"
+  if a.w = b.w then size a else failwith "bv binop widths differ"
 [@@inline always]
 
 let bind1 f a = create ~size:a.w (f ~size:a.w a.v) [@@inline always]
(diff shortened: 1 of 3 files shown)


lib/util/bitvec.ml
```diff
@@ -78,110 +79,82 @@ let of_string i =
...
 
 (* wrap signed bv operation *)
-let bind2_signed f a b =
-  size_is_equal a b;
-  create ~size:a.w (f (to_signed_bigint a) (to_signed_bigint b))
-[@@inline always]
+(*let bind2_signed f a b =
+  let size = size_is_equal a b in
+  create ~size (f ~size (to_signed_bigint a) (to_signed_bigint b))
+[@@inline always]*)
```
katrinafyi: delete unused

lib/util/bitvec.ml
```diff
@@ -78,110 +79,82 @@ let of_string i =
   { w; v }
 
 let size_is_equal a b = assert (size a = size b) [@@inline always]
-let bind1 f a = create ~size:a.w (f a.v) [@@inline always]
-let bind1_signed f a = create ~size:a.w (f a.v) [@@inline always]
+
+let size_if_equal a b =
+  if size a = size b then size a else failwith "bv binop widths differ"
```
katrinafyi:
```suggestion
  if size_is_equal a b then size a else failwith "bv binop widths differ"
```

lib/util/bitvec.ml
```diff
@@ -1,4 +1,5 @@
 open Containers
+module I = Fixed_width_arith
```
katrinafyi: i'd just inline `Fixed_width_arith` at all uses.  i can do that if u want. single letter module names always feel like functor parameters.

katrinafyi (approved):

lib/util/bitvec.ml
agle: its so long though

PR #183, opened 2026-06-18
Clean Aslp local var names and use `ID.generator`

katrinafyi:
This makes a new hashtable to associate each Aslp variable name with a counter. This also fixes the problem of dots inside variable names.

### DIFF
--- a/lib/transforms/aslp.ml
+++ b/lib/transforms/aslp.ml
@@ -392,15 +392,13 @@ struct
   let v_PSTATE_BTYPE : lexpr = Var.create "v_PSTATE_BTYPE" (Types.Bitvector 1)
 
   let v_BTypeCompatible : lexpr =
-    bincaml_local_var "v_BTypeCompatible" (Types.Bitvector 1)
+    Var.create "v_BTypeCompatible" (Types.Bitvector 1)
 
-  let v___BranchTaken : lexpr =
-    bincaml_local_var "v___BranchTaken" (Types.Bitvector 1)
-
-  let v_BTypeNext : lexpr = bincaml_local_var "v_BTypeNext" (Types.Bitvector 1)
+  let v___BranchTaken : lexpr = Var.create "v___BranchTaken" (Types.Bitvector 1)
+  let v_BTypeNext : lexpr = Var.create "v_BTypeNext" (Types.Bitvector 1)
 
   let v___ExclusiveLocal : lexpr =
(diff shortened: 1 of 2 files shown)


lib/transforms/aslp.ml
```diff
@@ -1,6 +1,37 @@
...
+
+  (** Returns a fresh ID. *)
+  let fresh { names; count } =
+    let prev = !count in
+    count := prev + 1;
+    prev
+
+  (** Returns the ID associated with the given name, or creates and stores a new
+      ID if the name is not yet known. *)
+  let find name st =
```
agle: ID.generator.decl_or_get

lib/transforms/aslp.ml
```diff
@@ -1,6 +1,37 @@
...
+module Aslp_IDs = struct
+  type t = {
+    names : (string, int) Hashtbl.t;
+        (** Names and their associated unique ID. *)
+    count : int ref;
+        (** Count of unique IDs {i ever} emitted by this generator. Importantly,
+            this counts names which have been forgotten. *)
+  }
+
+  let create () = { names = Hashtbl.create 16; count = ref 0 }
```
agle: Use ID.make_gen () or Fix.Gensym.create

lib/transforms/aslp.ml
```diff
@@ -1,6 +1,37 @@
...
+  let find name st =
+    match Hashtbl.find_opt st.names name with
+    | None ->
+        let n = fresh st in
+        Hashtbl.replace st.names name n;
+        n
+    | Some n -> n
+
+  (** Returns a copy of the {!t} which forgets all known name associations. *)
+  let forget st = { names = Hashtbl.copy st.names; count = ref !(st.count) }
```
agle: Use a separate map for substitutions, in the lifter state; their lifetime/scope (instruction) is different to that the generator (procedure).

Also the comment is wrong it copies the name associations.

katrinafyi: you're so right

katrinafyi:
I think I've got to most of the comments ([diff](https://github.com/agle/bincaml/compare/4e3893eb9550ca4868dddeffd10431721a31d855...7da190ac)). You can now inherit the block and local generators from a procedure, and they're just stored as `unit -> int` in the lifter state because that's all we need. As you suggest, the names are just stored in a separate hashmap adjacent to the generators and this works quite nicely.

agle (approved):

PR #184, opened 2026-06-19
Narrowing

b-paul:
Adds support for abstract domains with a narrowing operator. This is done by running the absint solver twice, the first iteration widening and the second iteration replacing the widening function with the narrowing function (this is how narrowing is presented in Mine's book).

This involved some moving around of a couple of things and having duplicated lattice signatures, which isn't very nice. The alternative is to have every abstract domain have a narrowing operator, which is possible (a default implementation is `a narrow b = a`) but I wasn't sure if this was an ideal solution and would like to know what others think.

### DIFF
--- a/lib/analysis/dataflow_graph.ml
+++ b/lib/analysis/dataflow_graph.ml
@@ -395,13 +395,6 @@ module type DFAnalysis = sig
   include Lattice_types.StateDomain with type t := t with type key_t = Var.t
 end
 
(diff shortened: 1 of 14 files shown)


agle:
We can combine the analysis domains to have a narrowing operator on everything and just define as identity if we don't want to use narrowing solver.

agle (approved):

PR #190, opened 2026-06-23
tree-sitter: more strict about deleting temporary `src` folder

katrinafyi:
might fix:

    Error when generating parser

    Caused by:
        Failed to write grammar.json to /home/alicia/Documents/programming/bincaml3/_build/default/tree-sitter/src -- Permission denied (os error 13)


katrinafyi:
Postponing for now, I guess it isn't a problem anymore. Can look into it again if the issue re-occurs.

PR #192, opened 2026-06-24
Lift calls to intrinsics

### LINKED ISSUE #73: Memory Allocation Intrinsics

agle:
fix: https://github.com/agle/bincaml/issues/73

closes https://github.com/agle/bincaml/pull/163

### DIFF
--- a/lib/lang/block.ml
+++ b/lib/lang/block.ml
@@ -125,11 +125,11 @@ let append_stmts (b : ('v, 'e) t) (stmt : ('v, 'v, 'e) Stmt.t list) :
     ('vv, 'ee) t =
   fmap_stmts_copy (fun v -> Vector.append_list v stmt) b
 
-let flat_map ~phi f (b : ('v, 'e) t) : ('vv, 'ee) t =
+let flat_map ?(rev = false) ~phi f (b : ('v, 'e) t) : ('vv, 'ee) t =
+  let it = if rev then Vector.to_iter_rev else Vector.to_iter in
+  let ot e = (if rev then Iter.rev e else e) |> Vector.of_iter in
   {
-    stmts =
-      Vector.to_iter b.stmts |> Iter.flat_map f |> Vector.of_iter
-      |> Vector.freeze;
+    stmts = it b.stmts |> Iter.flat_map f |> ot |> Vector.freeze;
     phis = phi b.phis;
     attrib = b.attrib;
   }
--- a/lib/lang/program.mli
+++ b/lib/lang/program.mli
@@ -74,6 +74,8 @@ val global_constants : t -> Var.t Iter.t
 val get_decl_by_name_id : string -> t -> (ID.t * declaration) option
 val get_decl_by_name : string -> t -> declaration option
 val get_proc_by_name : string -> t -> proc
+val get_decl : ID.t -> t -> declaration option
+val get_proc : ID.t -> t -> proc
 val get_implicit_decl_by_name : string -> t -> implicit_declaration option
 val declare_name : string -> t -> ID.t
 val declare_name_exn : string -> t -> ID.t
--- a/lib/lang/stmt.ml
+++ b/lib/lang/stmt.ml
@@ -24,12 +24,12 @@ module Intrinsic = struct
     | "@_malloc" -> Some Malloc
     | "@_calloc" -> Some Calloc
     | "@_free" -> Some Free
-    | "@_allcoa" -> Some AllocStack
+    | "@_alloca" -> Some AllocStack
     | "@_free_alloca" -> Some FreeStack
     | "@_aarch64_eval" -> Some Aarch64Eval
     | _ -> None
 
-  let pretty e =
+  let to_string e =
     (match e with
       | Havoc -> "@_havoc"
       | Malloc -> "@_malloc"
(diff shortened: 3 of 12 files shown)


lib/lang/program.mli
```diff
@@ -74,6 +74,8 @@ val global_constants : t -> Var.t Iter.t
 val get_decl_by_name_id : string -> t -> (ID.t * declaration) option
 val get_decl_by_name : string -> t -> declaration option
 val get_proc_by_name : string -> t -> proc
+val get_decl : ID.t -> t -> declaration option
+val get_proc : ID.t -> t -> proc
 val get_implicit_decl_by_name : string -> t -> implicit_declaration option
```
katrinafyi, line 79: there seems to be some inconsistency here about whether option is returned vs exception thrown. but no changes needed right now.

lib/lang/stmt.ml
```diff
@@ -24,12 +24,12 @@ module Intrinsic = struct
...
     | "@_calloc" -> Some Calloc
     | "@_free" -> Some Free
-    | "@_allcoa" -> Some AllocStack
+    | "@_alloca" -> Some AllocStack
     | "@_free_alloca" -> Some FreeStack
     | "@_aarch64_eval" -> Some Aarch64Eval
     | _ -> None
 
-  let pretty e =
+  let to_string e =
```
katrinafyi: why the renaming? this does not return a `string`.

lib/transforms/aarch64_intrin.ml
```diff
@@ -0,0 +1,67 @@
...
+      names =
+        [
+          ProcName "@alloca";
+          ProcName "@_alloca";
+          NameAttrib "alloca";
+          HasAttrib (StringMap.singleton ".isintrin" (`String "alloca"));
+        ];
+      in_args = [ "$mem"; "R0_in" ];
+      out_args = [ "$mem"; "R0_out" ];
+      intrin = Free;
```
katrinafyi: shold be alloca

lib/transforms/intrin.ml
```diff
@@ -0,0 +1,103 @@
+open Lang
+open Common
+open Stmt.Intrinsic
+
+(** How procedures are selected *)
+type name =
+  | ProcName of string
+  | NameAttrib of string
+  | HasAttrib of Attrib.attrib_map
```
katrinafyi, line 9: I'd rename the type to `name_filter` or smth.

lib/transforms/intrin.ml
```diff
@@ -0,0 +1,103 @@
...
+  in_args : string list;  (** captures & modifies & in params to use in order *)
+  out_args : string list;
+      (** modifies (must be in-params also) & out params to use in order *)
+  intrin : Stmt.Intrinsic.t;  (** the intrin to generatre *)
+}
+(** Specification for replacement of functions by intrinsics *)
+
+open struct
+  module LOM = List.Traverse (Option)
+end
```
katrinafyi, line 24: for list of option in particular, you can just use https://c-cube.github.io/ocaml-containers/containers/CCList/index.html#val-all_some instead of building the traverse machinery.

lib/transforms/intrin.ml
```diff
@@ -0,0 +1,103 @@
...
+      let oa = Procedure.attrib proc in
+      StringMap.for_all
+        (fun k v -> StringMap.find_opt k oa |> Option.exists (Attrib.equal v))
+        a
+    in
+    match name with
+    | ProcName n -> String.equal n (ID.to_string (Procedure.id proc))
+    | NameAttrib name ->
+        has_attrib (StringMap.singleton name_attrib_key (`String name))
+    | HasAttrib a -> has_attrib a
```
katrinafyi: move out of function body

lib/passes.ml
```diff
@@ -48,6 +48,16 @@ module PassManager = struct
...
     }
 
+  let lift_intrinsics_aarch64 =
+    {
+      name = "lift-intrinsics-aarch64";
+      apply = Prog Transforms.Aarch64_intrin.transform;
+      doc =
+        "Lift procedure calls to matched intrinsics in aslp/aarch64 abi to \
+         intrinsic calls";
+      invariants = Invariants.establishes ~invalidates:[] [ Params ];
```
katrinafyi: should this establish the general params property?

lib/lang/block.ml
```diff
@@ -125,11 +125,11 @@ let append_stmts (b : ('v, 'e) t) (stmt : ('v, 'v, 'e) Stmt.t list) :
     ('vv, 'ee) t =
   fmap_stmts_copy (fun v -> Vector.append_list v stmt) b
 
-let flat_map ~phi f (b : ('v, 'e) t) : ('vv, 'ee) t =
+let flat_map ?(rev = false) ~phi f (b : ('v, 'e) t) : ('vv, 'ee) t =
+  let it = if rev then Vector.to_iter_rev else Vector.to_iter in
+  let ot e = (if rev then Iter.rev e else e) |> Vector.of_iter in
```
katrinafyi, line 130: what's the purpose and intention of the rev flag? seems subtle and risky

katrinafyi (approved):
looks ok in principle. i would like to see more separation of concerns around the intrinsic matching -> transforming -> abi process, but will not block on this issue and can be refactored in future.

lib/lang/block.ml
agle: To reverse the iteration order, I just didn't want to make a separate function

katrinafyi: ok. just checking, it's not used in this PR functionality and it's just an extra for now?

if f returns multiple values, those will also be reversed. is that intentional? in any case, doc comment :)

agle: Its for consistency because its used by map_stmts_topo_rev which this pr adds because it needs map_stmts_top_fwd to match the iter_block_topo_fwd and iter_blocks_topo_rev in procedure.

agle: Yes it would be far more unexpected to not reverse the output stmts order...

PR #193, opened 2026-06-24
Aslp CFG construction using diamond ADT and zippers

katrinafyi:
Closes https://github.com/agle/bincaml/pull/191

- [x] TODO maybe: fix up the docs after upside down

### DIFF
--- a/lib/transforms/aslp/bincaml_ibi_make.ml
+++ b/lib/transforms/aslp/bincaml_ibi_make.ml
@@ -14,7 +14,7 @@ struct
   type lexpr = Aslp_lexpr.t
   type stmt = Aslp_state.stmt
   type branch = [ `T | `F | `M ]
-  type ast = Aslp_state.aslp_block Diamond.diamond
+  type ast = Aslp_state.aslp_diamond
 
   (** {2 Bincaml-specific utility functions} *)
 
@@ -199,13 +199,11 @@ struct
    fun cond ->
     let st = !bincaml_lifter_state and ncond = Expr.BasilExpr.boolnot cond in
 
-    let mk assume = Diamond.empty (Aslp_state.empty_block ?assume ()) in
-    let left = mk (Some cond) and right = mk (Some ncond) in
+    let left = Diamond.empty (Aslp_state.empty_block ~assume:cond ())
+    and right = Diamond.empty (Aslp_state.empty_block ~assume:ncond ()) in
 
-    let diamond =
-      st.diamond
-      |> Diamond.append_diamond ~left ~right ~value:(Aslp_state.empty_block ())
-    in
+    let value = Aslp_state.empty_block () in
+    let diamond = st.diamond |> Diamond.append_diamond ~left ~right ~value in
     bincaml_lifter_state := { st with diamond };
     (`T, `F, `M)
 
--- a/lib/transforms/aslp/bincaml_ibi_make.ml
+++ b/lib/transforms/aslp/bincaml_ibi_make.ml
@@ -3,7 +3,7 @@ open Common
 
 (** Makes a concrete module which implements {!Bincaml_ibi.IBI}. *)
 module Make (S : sig
-  val bincaml_lifter_state : Aslp_state.lifter_state ref
+  val initial_lifter_state : Aslp_state.lifter_state
 end) =
 struct
   (** {2 Type definitions} *)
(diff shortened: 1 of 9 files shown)


agle:
Does this make #191 redundant?

lib/transforms/aslp/diamond.ml
```diff
@@ -0,0 +1,272 @@
...
+    concerned with what happens at the {i end} of a control-flow diamond. The
+    "upside-down" view makes the control-flow join values conveniently
+    accessible. *)
+type 'a diamond =
+  | Leaf of 'a
+  | Diamond of {
+      pred : 'a diamond;
+      left : 'a diamond;
+      right : 'a diamond;
+      value : 'a;
```
agle, line 41: What value? naming

lib/transforms/aslp/bincaml_ibi_make.ml
```diff
@@ -199,40 +202,29 @@ struct
...
-    (`T t, `F f, `M (m, (t, f)))
+    let left = Diamond.empty (Aslp_state.empty_block ~assume:cond ())
+    and right = Diamond.empty (Aslp_state.empty_block ~assume:ncond ())
+    and value = Aslp_state.empty_block_unconditional () in
+
+    let diamond = st.diamond |> Diamond.append_diamond ~left ~right ~value in
+    bincaml_lifter_state := { st with diamond };
+    (`T, `F, `M)
 
   let f_switch_context : branch -> unit =
```
agle, line 215: Doesn't f_switch_context usually allow moving to an arbitrary point in the control flow? It seems to only let you move within a single branch, or are is that an assumption we can make of the generated lifter?

We've established that despite the given structure in the offline transform itself, below, subsequent passes seem to allow it to emit code which generates branches in different orders (image). Hence this needs a more sophisticated approach to switching append context.

> [aslp/libASL/offline_transform.ml at 92e4fd949402056e16e3747e66922060be57ecc8 · UQ-PAC/aslp](https://github.com/UQ-PAC/aslp/blob/92e4fd949402056e16e3747e66922060be57ecc8/libASL/offline_transform.ml#L807-L819)

image

lib/transforms/aslp/diamond.ml
```diff
@@ -0,0 +1,270 @@
+(** A zipper is a concept in functional programming (especially, Haskell).
```
agle: What this file does is more important than the concept of zippers. This should be an implementation detail with the headline comment being a description of cfg diamonds.

agle (commented):
Leaving further review for later;

It seems like a lot of complexity for something that was relatively simple in the scala implementation; https://github.com/UQ-PAC/BASIL/blob/716ec6d2e2c19aa37ab5625b5a824ee2de32117b/src/main/scala/translating/offlineLifter/OfflineLifter.scala#L26-L107 notwithstanding pc injection

Maybe it would be easier for IBI to write to a simple to a block sequence and extract to a diamond structure for injecting the PC? They somewhat seem to want different structures.

lib/transforms/aslp/diamond.ml
katrinafyi: It's meant to be generic to the kind of value. If it helps, I could name it something like "join" which is based on its relative position inside a diamond. But other functions will still make reference to "a focused value", for example, and I don't know how that could be changed.

lib/transforms/aslp/bincaml_ibi_make.ml
katrinafyi: Yeah. I have a plan but I'd like to do it in a follow-up PR, so this PR focuses on the foundation of the diamond structure. If you could review this PR without looking at any of the context switching, that would be appreciated.

lib/transforms/aslp/diamond.ml
katrinafyi: Yeah

katrinafyi (commented):
Fundamentally, the control structure is diamonds inside diamonds. I disagree that PC analysis and IBI need different structures, and I don't see how a diamond ADT is a bad structure for either PC propagation or IBI implementation - aside from the thing about context switching, and that can be managed. To design the whole CFG around having a convenient branch ID is not a good idea.

I'm not ignorant to these tradeoffs. I am making technical decisions while giving consideration to the issues you mention, as well as the issues and requirements that I foresee from experience.

I think the diamond ADT is the most sensible representation. I think this is clear in the aslp_state.ml diff - all of the code shuffling gotos and entry/exit IDs is gone. It might be unconventional, but we should not default to "obvious" implementations when better ones exist. This is the whole reason we are here, in bincaml. It's why use we Ocamlgraph, for instance, and why we don't copy BASIL wholesale.

I'm open to iterating and refining the presentation, but I'm not going to throw it out because of tiny hiccups. That said, I'm also not obstinate. If we run into a big enough roadblock, be assured that I am capable of throwing it away. We haven't got anywhere close to that point yet.

agle (approved):
I haven't made any criticism of the diamond/zipper situation beyond it implementing context switching wrong and being a lot of code & explanation.  Skimming the implementation and your code comments I can't see any problems; its simple enough.

PR #195, opened 2026-06-26
add explicit menhir dependency to capstone to avoid dune differences

katrinafyi:
this is not great because capstone doesn't need menhir, but i guess recent dune versions insist on adding a menhir dependency (with automatic version constraint) to every package in thr project.

specifying it explicitly avoids different behaviour in different dune versions.

### DIFF
--- a/capstone_arm64_disas.opam
+++ b/capstone_arm64_disas.opam
@@ -13,6 +13,7 @@ depends: [
   "ocaml"
   "ppx_expect"
   "odoc" {with-doc}
+  "menhir" {>= "20180523"}
 ]
 build: [
   ["dune" "subst"] {dev}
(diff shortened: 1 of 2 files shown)


agle:
What is the observable difference in behaviour? If you're installing capstone purely because its a dependency of bincaml it should always use the explicit version right?

katrinafyi:
Menhir isn't a real dependency of the capstone package. This PR was to work around this bug https://www.github.com/ocaml/dune/issues/14428, but since the bug has been fixed and a new version of dune is available, this PR probably isn't needed anymore. We should just avoid using 3.23 and be aware that that dune version can cause this problem.

katrinafyi:
Closing not needed. Cherry-pick https://github.com/agle/bincaml/commit/e9d698671b814472cc8fb0e89001ed1d069496f6 if we want it.

PR #197, opened 2026-06-26
Fix wrapped interval join soundess bug

b-paul:
it seems that the "if Top or" part of this algorithm was forgotten :(

image

this was only caught because CI randomly failed after generating a very rare expression that hit a very specific equality edge case in the sdiv logic `when not ((au = smin && bl = neg1) || (al = smin && bu = neg1))` that happened to create a weird join shape. it is worrying to think that other edge cases like this could be lurking in the depths...


agle (approved):
Very cool---its always sdiv

PR #200, opened 2026-06-29
Aslp frontend

katrinafyi:
This implements the Aslp frontend. It's a bit big.

- all arm variables are added all of the time.
- PC assignments are forwarded into an assign at the merge point, immediately before the nondeterministic goto. is this enough or should we put it right into the assume of the successors?
- todo update bincaml-test with new .address attribute

See tests at bottom half of https://github.com/agle/bincaml/blob/big-aslp/test/transforms/test_aslp.ml.

Here is an example for `b.lt #0x10`. You can see the propagated PC ITE assignment in the block_3. 
```ocaml
    proc @Sqrt()  -> () {  }
      captures $PC:bv64

    [
       block %Sqrt_code_4 { .address = 4196316 } [
         assume eq(0x4007dc:bv64, $PC);
         goto (%block);
       ];
       block %block { .asm = "b.lt #0x10" } [
         guard true;
         goto (%block_2,%block_1);
       ];
       block %block_1 [
         guard boolnot(eq($PSTATE_N, $PSTATE_V));
         var BranchTaken:bool := true;
         $PC:bv64 := 0x4007ec:bv64;
         goto (%block_3);
       ];
       block %block_2 [
         guard boolnot(boolnot(eq($PSTATE_N, $PSTATE_V)));
         (var BranchTaken:bool := false, $PC:bv64 := 0x4007e0:bv64);
         goto (%block_3);
       ];
       block %block_3 [
         guard true;
         $PC:bv64 := if boolnot(eq($PSTATE_N, $PSTATE_V)) then 0x4007ec:bv64 else 0x4007e0:bv64;
         assert boolor(eq(0x4007fc:bv64, $PC), eq(0x4007e0:bv64, $PC));
         goto (%Sqrt_code_3,%Sqrt_code);
       ];
       block %Sqrt_code [ assume eq(0x4007e0:bv64, $PC); goto (%ret); ];
       block %Sqrt_code_3 [ assume eq(0x4007fc:bv64, $PC); goto (%ret); ];
       block %ret [ return; ]
    ];
```

Closes https://github.com/agle/bincaml/pull/194
Closes https://github.com/agle/bincaml/pull/198
Closes https://github.com/agle/bincaml/pull/199


katrinafyi:
The submodule reference is set to a commit which is not in bincaml-test main atm.

lib/transforms/aslp/aslp.ml
```diff
@@ -7,31 +7,233 @@
...
+      These [Some] values are returned in the first tuple element. Upon reaching
+      a value which yields [None], that value and values after it are returned
+      in the second tuple element. *)
+  let rec take_drop_while_map f = function
+    | [] -> ([], [])
+    | hd :: rest as all -> (
+        match f hd with
+        | Some x -> (
+            match take_drop_while_map f rest with a, b -> (x :: a, b))
+        | None -> ([], all))
```
agle, line 51: I wouldn't object to moving this generic helper to Common.List extending CCList

lib/transforms/aslp/aslp.ml
```diff
@@ -7,31 +7,233 @@
...
+    an {!Lang.Stmt.Intrinsic.Aarch64Eval} intrinsic call. Otherwise, returns
+    [None].
+
+    Raises an exception if an {!Lang.Stmt.Intrinsic.Aarch64Eval} intrinsic call
+    has an unexpected structure. *)
+let aarch64_intrin_of_stmt :
+    Program.stmt -> (Bitvec.t * Attrib.attrib_map) option = function
+  | Stmt.Instr_IntrinCall { attrib; lhs; name = Aarch64Eval; args } -> (
+      match (lhs, args) with
+      | [], [ E (Constant { const = `Bitvector op }) ] -> Some (op, attrib)
```
agle, line 66: I think the intrin should be explicitly passed the instruction address---this makes the address the concern of the block translator (`gfir_to_ir.ml`) where it already has direct access to the absolute address, rather than recovering it from the more sparse information passed  + assumption the intrinsics inherently have an address relative to the block theyre in which is just a coincidental result of the frontend.

lib/transforms/aslp/aslp.ml
```diff
@@ -7,31 +7,233 @@
...
+  | Some (Variable { binding }) -> binding
+  | _ -> failwith "aarch64_mem_of_prog: no $mem found"
+
+(** Returns the byte address of the given block, if present. *)
+let address_of_block block =
+  block.Lang.Block.attrib
+  |> StringMap.find_opt ".address"
+  |> Option.map (function
+    | `Integer x -> x
+    | _ -> failwith "address_of_block: invalid type in .address")
```
agle: Attrib.find_int_opt %> Option.get_exn_or "no integer address field"

lib/transforms/aslp/aslp.ml
```diff
@@ -7,31 +7,233 @@
...
+    If there are no Aarch64 statements, this still succeeds but will return [[]]
+    in the second and third elements. *)
+let partition_aarch64_stmts stmts =
+  let before, stmts =
+    List.take_drop_while (Option.is_none % aarch64_intrin_of_stmt) stmts
+  in
+  let intrins, after =
+    Internal.take_drop_while_map aarch64_intrin_of_stmt stmts
+  in
+  (before, intrins, after)
```
agle, line 82: I could imagine a version of this function that returns an alternating sequence of decode ops and regular ops; this would support running the transform on  a program with only a few decode intrinsics rather than explicitly the output of gtirb loader ; cf. `List.group_succ `. It should still produce a sequence of diamonds with which to replace the block.

lib/transforms/aslp/aslp.ml
```diff
@@ -7,31 +7,233 @@
...
+    external control flow using the returned IDs. *)
+let insert_one_diamond ~proc dia =
+  let next_id _ = ID.fresh ~name:"%block" (Procedure.block_ids proc) () in
+  let with_ids = dia |> Diamond.enumerate_with_successors next_id in
+
+  let proc =
+    with_ids |> Diamond.iter_backwards
+    |> Iter.fold
+         (fun proc (id, successors, lifter_block) ->
+           let { Aslp_state.stmts; assume } = lifter_block in
```
agle: Is this `Aslp_state` a local open for the block type? Confusing

lib/transforms/aslp/aslp.ml
```diff
@@ -7,31 +7,233 @@
...
+          Stmt.Instr_Assign { attrib = Attrib.empty; al } :: after
+      | Some (Leaf _) | None -> after
+    in
+
+    (* Append back things which were previously after the Aarch64 intrinsic calls,
+     but append them to the *last* block. Finally, fix up successors. *)
+    let proc =
+      Procedure.modify_block proc last (fun b -> Block.append_stmts b after)
+    in
+    Procedure.replace_block_succs proc last block_successors)
```
agle, line 187: Could be generalized by adding function to procedure `add_block_succ` or `add_block_pred` which injects an empty block after/before the block and replaces the successors/predecessors with the original. Then the diamond can be injected between the original block and this injected block without thinking about the original edges.

lib/transforms/aslp/aslp.ml
```diff
@@ -7,31 +7,233 @@
...
-let lift_code_block (module I : Bincaml_ibi.IBI) ~address opcodes =
-  opcodes
-  |> Iter.mapi (fun i op ->
+    Requires and ensures that the IBI is in the "reset" state. *)
+let lift_code_block (module I : Bincaml_ibi.IBI) ~address =
+  List.mapi (fun i op ->
       let address = Bitvec.add (Bitvec.create ~size:64 Z.(~$4 * ~$i)) address in
       lift_opcode (module I) ~address op)
-  |> Iter.to_list
+
```
agle, line 37: I think the intrin should be explicitly passed the instruction address when its created---this makes the address the concern of the block translator (`gfir_to_ir.ml`) where it already has direct access to the absolute address, rather than recovering it from the block address  + assumption the intrinsics inherently are sequential instructions with an address relative to the block theyre in.

lib/transforms/aslp/aslp_lexpr.ml
```diff
@@ -77,3 +78,34 @@ let to_var x =
...
+
+let predefined =
+  [
+    PC; SP_EL0;
+
+    R (Some 0); R (Some 1); R (Some 2); R (Some 3); R (Some 4); R (Some 5); R (Some 6);
+    R (Some 7); R (Some 8); R (Some 9); R (Some 10); R (Some 11); R (Some 12);
+    R (Some 13); R (Some 14); R (Some 15); R (Some 16); R (Some 17); R (Some 18);
+    R (Some 19); R (Some 20); R (Some 21); R (Some 22); R (Some 23); R (Some 24);
+    R (Some 25); R (Some 26); R (Some 27); R (Some 28); R (Some 29); R (Some 30);
```
agle: Could we use `Iter.int_range ~from:0 ~to:29 |> Iter.map (fun i -> R (Some i)) |> Iter.to_list` ?

lib/transforms/aslp/diamond_zipper.ml
```diff
@@ -0,0 +1,304 @@
...
+    {{:http://strictlypositive.org/diff.pdf} McBride}.
+
+    This module was mostly implemented by following the
+    {{:https://wiki.haskell.org/Zipper} Haskell wiki} (if reading this, be aware
+    that our inner nodes store values). If you enjoy Tony Morris's teaching
+    style, he has
+    {{:https://www.youtube.com/watch?v=HqHdgBXOOsE} a talk on YouTube} (the
+    first 10 minutes are most relevant). *)
+
+open Diamond
```
agle: repeated open

lib/transforms/aslp/diamond.ml
```diff
@@ -42,242 +42,117 @@ type 'a diamond =
...
+(** Maps the given function over the {!diamond} structure, processing values
+    {i backwards} in control-flow order. *)
+let rec map_backwards f = function
+  | Leaf x -> Leaf (f x)
+  | Diamond { value; left; right; pred } ->
+      let value = f value in
+      let left = map_backwards f left
+      and right = map_backwards f right
+      and pred = map_backwards f pred in
+      Diamond { value; left; right; pred }
```
agle, line 84: Map backwards and forwards are identical modulo side-effects of `f`? Maybe they should be called iter_map or something?

Actually it looks like this is dead code.

lib/transforms/aslp/bincaml_ibi_make.ml
```diff
@@ -251,7 +262,7 @@ struct
 
   let f_gen_store : lexpr -> expr -> unit =
    fun lhs rhs ->
-    bincaml_emit
+    bincaml_internal_emit
```
agle, line 257: In basil we generate a clear op for every local variable declaration so that we don't have uninitialised variables appearing on the RHS when the vector instructions incrementally initialise the variable.

agle (approved):
LGTM

lib/transforms/aslp/aslp.ml
katrinafyi: I'm hesitant to do because we already use containers in a stdlib-shadowing way, so to look for API docs you already have to go to 2 places. Extra bincaml-specific extensions would be yet another place to look from. 

but we can certainly promote it outside of this aslp module somewhere. i wont change anything right now.

lib/transforms/aslp/aslp.ml
katrinafyi: Yeah. This would also let the transform work at the level of per intrinsic rather than at the block level. Will look into as part of integration work.

lib/transforms/aslp/aslp.ml
katrinafyi: tq. ended up being slightly clunky with
```ocaml
  Some (`Assoc block.Lang.Block.attrib)
  |> Attrib.find_int_opt ".address"
```
but it's a bit neater and can improve later

lib/transforms/aslp/aslp.ml
katrinafyi: acknowledge

lib/transforms/aslp/aslp.ml
katrinafyi: i guess it's a kind of local open to disambiguate the record `stmts` field. changed to
```ocaml
           let { stmts; assume } : Aslp_state.aslp_block = lifter_block in
```

lib/transforms/aslp/aslp.ml
katrinafyi: hmm yeah. there's also the problem that manipulating succs/preds doesn't interact nicely work entry/return edges (which leads to one assert). i probably wont add block manipulating functions myself but wil think about the problem

lib/transforms/aslp/aslp.ml
katrinafyi: yeah. future

lib/transforms/aslp/aslp_lexpr.ml
katrinafyi: will do

lib/transforms/aslp/diamond_zipper.ml
katrinafyi: good spot

lib/transforms/aslp/diamond.ml
katrinafyi: `map_forwards` is used by enumerate to generate the block IDs and as long as there's one direction i'll keep the other. changed comment to mention side effects

lib/transforms/aslp/bincaml_ibi_make.ml
katrinafyi: will do

lib/transforms/aslp/aslp.ml
agle: Hm we generate docs for all the deps so I hope it wouldnt be too bad to just look here https://agle.github.io/bincaml/bincaml/bincaml.util/Bincaml_util/Common/index.html#module-List

katrinafyi: i think it would be a little bit bad. i personally have never gone there and always go to c-cube website

PR #202, opened 2026-07-01
gtirb: preserve conditional fallthrough edges

### LINKED ISSUE #201: Anomaly in Gtirb conditional branch successor addresses
See code below and note the successors. This is in gtirb.t.

The address of the beginning of the `b.lt #0x10` is `0x4007a0 + 4 * 12 = 0x4007d0`, so its fallthrough PC is `0x4007d4` but this is not in the succesors.

Also, if you have a conditional branch, it either increments by the constant (if branched) or increments by 4 (if fallthrough). So, here, the difference between the two possible successor addresses should be `0x10 - 4 = 12`. However, the two successors actually have a difference of 28.

I guess that this could be a case of ddisasm resolving/inferring too much? Or, there could be a bug. This issue could cause failing PC assertions.

This does not affect all blocks. `Sqrt_code_2` and `Sqrt_code_1` are both okay, both of those being terminated by `b`. `Sqrt_code_5` also seems okay (though it only has one successor for a `b.ne`?).

```ocaml
block %Sqrt_code_4 { .address = 4196256;
    .gtirb_block = "rCSSdLZcRB2TKAu9h+WCqg";
    .succ = [ { .address = 4196348; .target = "internal:lr6o4ptnRiK3TGR1gpiWGg" };
        { .address = 4196320; .conditional = "true"; .direct = "true";
            .target = "internal:HVqN0/3+RWiLPKsHRvUqeg"; .type = "Type_Branch" } ] } [
  assume eq(0x4007a0:bv64, $PC);

  // 12 normal opcodes
  call @_aarch64_eval(0xf94017e1:bv32) { .asm = "ldr x1, [sp, #0x28]" };
  call @_aarch64_eval(0xf94013e0:bv32) { .asm = "ldr x0, [sp, #0x20]" };
  call @_aarch64_eval(0x8b000020:bv32) { .asm = "add x0, x1, x0" };
  call @_aarch64_eval(0xd37ffc01:bv32) { .asm = "lsr x1, x0, #0x3f" };
  call @_aarch64_eval(0x8b000020:bv32) { .asm = "add x0, x1, x0" };
  call @_aarch64_eval(0x9341fc00:bv32) { .asm = "asr x0, x0, #1" };
  call @_aarch64_eval(0xb9001fe0:bv32) { .asm = "str w0, [sp, #0x1c]" };
  call @_aarch64_eval(0xb9401fe0:bv32) { .asm = "ldr w0, [sp, #0x1c]" };
  call @_aarch64_eval(0x1b007c00:bv32) { .asm = "mul w0, w0, w0" };
  call @_aarch64_eval(0x93407c00:bv32) { .asm = "sxtw x0, w0" };
  call @_aarch64_eval(0xf94007e1:bv32) { .asm = "ldr x1, [sp, #8]" };
  call @_aarch64_eval(0xeb00003f:bv32) { .asm = "cmp x1, x0" };

  call @_aarch64_eval(0x5400008b:bv32) { .asm = "b.lt #0x10" };
  assert boolor(eq(0x4007fc:bv64, $PC), eq(0x4007e0:bv64, $PC));
  goto (%Sqrt_code_3,%Sqrt_code);
];
```

agle:
fix: #201

### DIFF
--- a/lib/gtirb/gfir_to_bincaml.ml
+++ b/lib/gtirb/gfir_to_bincaml.ml
@@ -150,18 +150,27 @@ let reorder_fallthrough (g : Gfir.G.t) =
       |> List.partition
            Gfir.Edge.(
              function
-             | _, Some { type' = Gfir.Edge.Type_Fallthrough }, v -> true
+             | _, Some { type' = Type_Fallthrough; conditional = false }, v ->
+                 true
              | _ -> false)
+    in
+    let is_call_edge = function
+      | Gfir.Edge.(_, Some { type' = Type_Call }, _) -> true
+      | Gfir.Edge.(_, Some { type' = Type_Syscall }, _) -> true
+      | _ -> false
     in
     (* check if these successors have no successors, this transform is made
          safe by the assertions *)
-    match ft with
-    | [] -> g
-    | [ ((_, _, fallthru_tgt) as fe) ] ->
-        let g = Gfir.G.remove_edge_e g fe in
-        List.fold_left
-          (fun g (_, _, s) -> Gfir.G.add_edge g s fallthru_tgt)
-          g nft
+    match (ft, nft) with
+    | [], _ -> g
+    | [ ((_, _, fallthru_tgt) as fe) ], os ->
+        if not @@ List.exists is_call_edge os then g
+        else begin
+          let g = Gfir.G.remove_edge_e g fe in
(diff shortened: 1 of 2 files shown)


katrinafyi (approved):
yeah it's good that this is an easily fixable problem. thanks for looking into it.

PR #203, opened 2026-07-03
Clean up Aslp BFS iteration a bit

katrinafyi:
To be honest, it is debatable whether this is easier to read than the old code.

I think it is slightly, because you don't need do so much bookkeeping of directions. It also makes the tree structure more explicit and adds a scan function.


agle (approved):
lgtm

PR #204, opened 2026-07-07
Add and use address argument in each Aslp intrinsic

katrinafyi:
headline changes:
- gtirb loader now attaches address to every intrinsic, and the transform looks for this. the transform now works one statement at a time.
- to help this, i've added some helper methods in Procedure for block control flow. i want advice on whether they look good.
- also added a "micro-pass" to compute stmt address from block .address to avoid needing address on every statement in tests, which would've been tedious. this has to be run manually if it's wanted

other new features:
- exceptions while lifting now just lead to a `.error` attribute on the intrinsic, rather than crashing. transforming continues with the next valid opcode. this can be seen in a test case.

other output changes:
- a final combined PC assignment is now emitted to the last block in the ASLp output (previously, it was just recorded internally but not used). let me know if this is enough for future guard cleanup, or whether it needs to be forwarded even more.
- guard trues are omitted.
- PC ITEs which are the same on both side are simplified to avoid the ITE.
- only used ARM global variables are generated.

however:
- transforming one statement at a time leads to overall `n^2` in the number of intrinsics, because we make new block stmt vectors after each intermediate intrinsic and it's cloned each time. i don't know how to improve this without introducing a new temporary block stmt list representation or something similarly hacky. edit: I have an idea now. next pr.


lib/gtirb/gfir_to_bincaml.ml
```diff
@@ -118,20 +118,23 @@ let add_new_code_block (all_blocks : block UUIDMap.t) temp_proc succ_addr
       let attrib = Option.fold Attrib.merge_map_shadow attrib attrib' in
       let instrs =
         opcodes
-        |> List.map (fun op ->
+        |> List.mapi (fun i op ->
             let op = Opcode.of_be_bytes op in
+            let asm =
+              if conf.disas then
+                [ (".asm", `String (CCResult.retract (Disas.dis_op op))) ]
```
agle: This should just be `Result` because we open Common.

lib/gtirb/gfir_to_bincaml.ml
```diff
@@ -118,20 +118,23 @@ let add_new_code_block (all_blocks : block UUIDMap.t) temp_proc succ_addr
...
                 name = Stmt.Intrinsic.Aarch64Eval;
                 args =
                   [ Expr.BasilExpr.const (`Bitvector (Opcode.to_bitvec op)) ];
-                attrib =
-                  op
-                  |> (if conf.disas then Disas.dis_op %> Result.to_option
-                      else const None)
-                  |> Option.map_or ~default:Attrib.empty (fun asm ->
-                      StringMap.singleton ".asm" (`String asm));
+                attrib = StringMap.of_list (asm @ address);
```
agle: I would rather it go in the argument list than the attribute as it defines the behaviour of the intrinsic.

lib/transforms/aslp/aslp.ml
```diff
@@ -35,65 +35,68 @@ let lift_code_block (module I : Bincaml_ibi.IBI) ~address =
...
     [None].
 
     Raises an exception if an {!Lang.Stmt.Intrinsic.Aarch64Eval} intrinsic call
     has an unexpected structure. *)
-let aarch64_intrin_of_stmt :
+let aarch64_intrin_of_stmt ?(include_failed = false) :
     Program.stmt -> (Bitvec.t * Attrib.attrib_map) option = function
-  | Stmt.Instr_IntrinCall { attrib; lhs; name = Aarch64Eval; args } -> (
+  | Stmt.Instr_IntrinCall { attrib; lhs; name = Aarch64Eval; args }
+    when include_failed || not (StringMap.mem error_attrib_key attrib) -> (
```
agle, line 51: Why would there be error attributes at this stage? Re-running on a program with failures? Seems like it should just throw again rather than have different behaviour.

Gosh never mind its to skip errors on the next call to `transform_one_statement`.

lib/transforms/aslp/aslp_util_internal.ml
```diff
@@ -0,0 +1,18 @@
+(** Returns [Some (hd x, tl x)] if [x] is non-empty, otherwise returns [None].
+*)
+let uncons = function [] -> None | hd :: tl -> Some (hd, tl)
```
agle, line 9: Should define unsnoc for completeness ;)

lib/transforms/aslp/aslp_util_internal.ml
```diff
@@ -0,0 +1,18 @@
...
+
+(** [span_while_some f xs] returns the longest prefix of [xs] where the elements
+    yield [Some] when mapped through [f].
+
+    These [Some] values are returned in the first tuple element. Upon reaching a
+    value which yields [None], that value and values after it are returned in
+    the second tuple element. *)
+let span_while_some f =
```
agle, line 17: This is unused

lib/transforms/aslp/aslp_state.ml
```diff
@@ -191,10 +191,44 @@ let ensure_pc_consistency ~address state =
...
+          |> Diamond_zipper.modify
+               (add_stmt_to_block ~stmt ~allow_double_pc:true)
+      | Some _, None | None, Some _ -> failwith "pcs should already agree")
+
+(** Returns an {!Lang.Stmt.Instr_Assume} statement for the given block's guard,
+    if it is non-trivial. *)
+let assume_of_aslp_block = function
+  | { assume = E (Constant { const = `Bool true }); _ } -> None
+  | { assume = body } ->
+      Some (Stmt.Instr_Assume { attrib = Attrib.empty; body; branch = true })
```
agle: This is instruction-internal control flow right? `branch` should probably be false unless there is a PC assignment (which should appear in the gtirb graph anyway). Not sure how we could validate that we have that level of leakage preservation though.

lib/transforms/aslp/aslp.ml
```diff
@@ -219,8 +218,38 @@ let transform_program prog =
...
+(** {1 Supplementary transformation} *)
+
+(** Micro-pass to insert [.address] on each {!Lang.Stmt.Intrinsic.Aarch64Eval}
+    intrinsic statement, computed from block-level [.address] attributes. This
+    can be useful in tests to avoid needing to type every address.
+
+    Existing statement [.address] attributes, if they exist, are unchanged and
+    do not interact with this transform. If a block has no [.address] attribute,
+    no changes are made to that block. *)
+let apply_stmt_addresses_from_block (block : _ Block.t) =
```
agle, line 225: Probably just put in the test file?

agle (commented):
Im not a huuuge fan of using the attributes to ensure progress in the face of lifter errors but I guess it works.

lib/transforms/aslp/aslp_util_internal.ml
katrinafyi: Unless you have strong objections, I'd like to keep it there. It could be useful for me in the future.

lib/transforms/aslp/aslp.ml
katrinafyi: Unless you have strong objections, I'd like to keep it there. It's closely related to the `aarch64_intrin_of_stmt` functions and may be useful for users or across multiple tests.

lib/gtirb/gfir_to_bincaml.ml
katrinafyi: It's true. I've moved it into the second argument in the list now.

katrinafyi (commented):
Thanks for review. I've made the changes which are small but have big diffs because of the address change.

I'm also going to leave the `.error` attribute as is for now, though yeah it's not the best.

lib/transforms/aslp/aslp.ml
```diff
@@ -35,65 +35,64 @@ let lift_code_block (module I : Bincaml_ibi.IBI) ~address =
...
+          [
+            Constant { const = `Bitvector op };
+            Constant { const = `Bitvector address };
+          ] )
+        when Bitvec.size op = 32 && Bitvec.size address = 64 ->
+          Some (op, address, attrib)
+      | _ ->
+          failwith
+            "invalid Aarch64Eval args. expected @_aarch64_eval(op:bv32, \
+             addr:bv64)")
```
agle, line 68: Or `@_aarch64_eval(op:bv32)` and provided `default_address`

agle (approved):

lib/transforms/aslp/aslp.ml
katrinafyi: True, but that's really just an internal thing which is only used through `apply_stmt_addresses_from_block` so I don't feel a need to advertise it super widely.

PR #205, opened 2026-07-08
Nix maintenance

mira-alford:
Just some maintenance of the flake to keep the package and devshell working and up to date. It was missing the examples submodule which was causing me some headache getting it to pass checks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ trace.json
 **/*.dot
 
 .vscode/
+.direnv
+.envrc
--- a/flake.nix
+++ b/flake.nix
@@ -1,5 +1,7 @@
 {
   inputs = {
+    self.submodules = true;
+
     nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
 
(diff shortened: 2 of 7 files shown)


katrinafyi (commented):
Historically, I thought that fetching submodules was kinda sketchy in Nix. So the hope was that the package would disable tests, but you could still run the tests locally in the Dev shell where you manually manage the submodule. Did you run into issues with this?

Also, just a quick question, what's the need for vendorimg containers? Could you add that as a comment so we know when to devendor it?

mira-alford:
@katrinafyi 

Containers on nixpkgs is outdated, which hasn't been a problem up until now luckily, but we have started using CCResult recently which was causing issues. Ill go add a comment.

I could run the tests locally just fine, i didn't realize we were avoiding the nix package running them. I can go and undo that if you want but It seems to be working with submodules just fine for the time being.

katrinafyi (approved):
Makes sense! No further changes needed.

Thanks for looking into it and updating the Nix!

PR #213, opened 2026-07-08
Unescape str literals in frontend

agle:


katrinafyi (approved):
cool function. we may have to keep in mind if our escapes differ from Ocaml's, but no big deal right now.

agle:
When it dumps strings it usually uses Printf.sprintf "%S" so I'm legally specifying our escape to be ocaml's escape

PR #214, opened 2026-07-08
Acyclic CFA reduction

mira-alford:
Implements the acyclic cfa reduction transform from some old snippet of nicks thesis. Basically just flattens an acyclic, deterministic procedure in a single block.
Cheats a bit because we already have an SSA transform so we can skip some of the algorithm, yay!

### DIFF
--- a/lib/transforms/cfa_reduction.ml
+++ b/lib/transforms/cfa_reduction.ml
@@ -1,24 +1,35 @@
+(**
+  A transform which reduces the CFA (control-flow automaton)
+  to a single edge, representing control flow through linear
+  if-then-else expressions rather than phi nodes.
+
+  The original algorithm assumes no SSA, however this does
+  as it simplifies it greatly. Also assumes the CFA is acyclic
+  and pure.
+  *)
+
 open Lang
 open Lang.Common
+open Expr
 
-(* Assumes program is in SSA, acyclic and pure. *)
 let construct_final_edge proc =
   (* Termination condition for each edge. *)
-  let term : (IDSet.elt, Var.t) Hashtbl.t = Hashtbl.create 30 in
+  let termination_condition : (IDSet.elt, Var.t) Hashtbl.t =
+    Hashtbl.create 30
+  in
   Procedure.fold_blocks_topo_fwd
     (fun final_edge id block ->
-      let preds = Procedure.blocks_pred proc id |> Iter.to_list in
-
       (* Compute reachability of this block. *)
       let reachable =
+        match Procedure.blocks_pred proc id |> Iter.to_list with
         (* Always reachable if no predecessors. *)
-        if List.is_empty preds then Expr.BasilExpr.boolconst true
+        | [] -> Expr.BasilExpr.boolconst true
         (* Otherwise, reachability is ANY of the predecessors reachability. *)
-          else
-          Expr.BasilExpr.applyintrin ~op:`OR
-            (preds
-            |> List.filter_map (fst %> Hashtbl.get term)
-            |> List.map Expr.BasilExpr.rvar)
+        | preds ->
+            Expr.BasilExpr.applyintrin ~op:`OR
+              (preds
+              |> List.filter_map (fst %> Hashtbl.get termination_condition)
+              |> List.map Expr.BasilExpr.rvar)
       in
 
       (* Create ITE statements from phi nodes. *)
@@ -26,15 +37,14 @@ let construct_final_edge proc =
         let al =
           block.phis
           |> List.map (function
-            | ({ lhs; rhs = hd :: tl } : Var.t Block.phi) ->
+            | ({ lhs; rhs = (_, hd_var) :: tl } : Var.t Block.phi) ->
+                (* Head id is unused as it is the negation of everything. *)
                 let ite =
                   List.fold_left
                     (fun acc (in_edge, var) ->
-                      Expr.BasilExpr.ifthenelse
-                        (Expr.BasilExpr.rvar @@ Hashtbl.find term in_edge)
-                        (Expr.BasilExpr.rvar var) acc)
-                    (Expr.BasilExpr.rvar @@ snd hd)
-                    tl
+                      let cond = Hashtbl.find termination_condition in_edge in
+                      BasilExpr.(ifthenelse (rvar @@ cond) (rvar var) acc))
+                    (BasilExpr.rvar @@ hd_var) tl
                 in
                 (lhs, ite)
             | _ -> failwith "Encountered phi node with no rhs.")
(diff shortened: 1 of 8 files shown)


mira-alford:
Just noticed the tacked on simplify at the end messes it up by inlining things in a way that messes up the guards? Removing the simplify pass for now.

lib/transforms/cfa_reduction.ml
```diff
@@ -0,0 +1,104 @@
+open Lang
+open Lang.Common
+
+(* Assumes program is in SSA, acyclic and pure. *)
+let construct_final_edge proc =
+  (* Termination condition for each edge. *)
+  let term : (IDSet.elt, Var.t) Hashtbl.t = Hashtbl.create 30 in
```
katrinafyi: term is too generic, a longer name like "termination_condition"  would be better.

lib/transforms/cfa_reduction.ml
```diff
@@ -0,0 +1,104 @@
...
+          block.phis
+          |> List.map (function
+            | ({ lhs; rhs = hd :: tl } : Var.t Block.phi) ->
+                let ite =
+                  List.fold_left
+                    (fun acc (in_edge, var) ->
+                      Expr.BasilExpr.ifthenelse
+                        (Expr.BasilExpr.rvar @@ Hashtbl.find term in_edge)
+                        (Expr.BasilExpr.rvar var) acc)
+                    (Expr.BasilExpr.rvar @@ snd hd)
```
katrinafyi: Local open of BasilExpr can avoid some repetition

```suggestion
                  List.fold_left
                    (fun acc (in_edge, var) ->
                      let cond = Hashtbl.find term in_edge in
                      Expr.BasilExpr.(ifthenelse (rvar cond) (rvar var) acc))
                    (Expr.BasilExpr.rvar @@ snd hd)
```

I also add a local variable because I like to keep things in the local open as small as possible.

lib/transforms/cfa_reduction.ml
```diff
@@ -0,0 +1,104 @@
...
+            |> List.filter_map (fst %> Hashtbl.get term)
+            |> List.map Expr.BasilExpr.rvar)
+      in
+
+      (* Create ITE statements from phi nodes. *)
+      let ites : Program.stmt =
+        let al =
+          block.phis
+          |> List.map (function
+            | ({ lhs; rhs = hd :: tl } : Var.t Block.phi) ->
```
katrinafyi: There's a subtlety here that the block ID in `hd` is not used and it's just assumed to be the negation of everything else. Is this worth a comment?

Regardless, pattern matching on `(_, hd_var)` instead of `hd` will make this more clear and avoid the `snd`.

lib/transforms/cfa_reduction.ml
```diff
@@ -0,0 +1,104 @@
...
+      (* Construct our termination condition by combining
+          initial reachability with any assumes along this edge. *)
+      let termination_cond =
+        Expr.BasilExpr.applyintrin ~op:`AND
+          ([ reachable ]
+          @ (Block.stmts_iter block
+            |> Iter.filter_map (function
+              | Stmt.Instr_Assume { body; branch = true } -> Some body
+              | _ -> None)
+            |> Iter.to_list))
```
katrinafyi: I would prefer to factor out this filter (and avoid the repetition in the later inverse filter) by using a partition.

For example
```ocaml
      let guard_expressions, non_guard_stmts =
        Block.stmts_iter block |> Iter.to_list
        |> List.partition_map_either (function
          | Stmt.Instr_Assume { body; branch = true } -> Left body
          | stmt -> Right stmt)
```

This will also simplify the final list concats in this function.

lib/transforms/cfa_reduction.ml
```diff
@@ -0,0 +1,104 @@
...
+            block.stmts |> Vector.to_list
+            |> List.filter (fun s ->
+                not
+                @@
+                match s with
+                | Stmt.Instr_Assume { branch = true } -> true
+                | _ -> false);
+            [ termination ];
+          ])
+    List.empty proc
```
katrinafyi:
```suggestion
    [] proc
```

lib/transforms/cfa_reduction.ml
```diff
@@ -0,0 +1,104 @@
...
+      let reachable =
+        (* Always reachable if no predecessors. *)
+        if List.is_empty preds then Expr.BasilExpr.boolconst true
+        (* Otherwise, reachability is ANY of the predecessors reachability. *)
+          else
+          Expr.BasilExpr.applyintrin ~op:`OR
+            (preds
+            |> List.filter_map (fst %> Hashtbl.get term)
+            |> List.map Expr.BasilExpr.rvar)
+      in
```
katrinafyi: Generally prefer pattern matching rather than explicit conditions. I've also inlined the `preds` definition because it's only used here.

```suggestion
      (* Compute reachability of this block. *)
      let reachable =
        match Procedure.blocks_pred proc id |> Iter.to_list with
        (* Always reachable if no predecessors. *)
        | [] -> Expr.BasilExpr.boolconst true
        (* Otherwise, reachability is ANY of the predecessors reachability. *)
        | preds ->
            Expr.BasilExpr.applyintrin ~op:`OR
              (preds
              |> List.filter_map (fst %> Hashtbl.get term)
              |> List.map Expr.BasilExpr.rvar)
      in
```

lib/transforms/cfa_reduction.ml
```diff
@@ -0,0 +1,104 @@
...
+    |> Iter.fold (fun acc id -> Procedure.remove_block acc id) proc
+  in
+  let out_proc, id = Procedure.fresh_block out_proc ~stmts:final_edge () in
+
+  (* Make this the entry and return block. *)
+  let out_proc = Procedure.set_entry_block out_proc id in
+  Procedure.PG.map_graph
+    (fun g ->
+      Procedure.G.add_edge g (Procedure.Vert.End id) Procedure.Vert.Return)
+    out_proc
```
katrinafyi: Just shadow `proc` to avoid accidentally re-using it.

```suggestion
  let proc =
    proc |> Procedure.iter_blocks |> Iter.map fst
    |> Iter.fold Procedure.remove_block proc
  in
  let proc, id = Procedure.fresh_block proc ~stmts:final_edge () in

  (* Make this the entry and return block. *)
  let proc = Procedure.set_entry_block proc id in
  Procedure.PG.map_graph
    (fun g ->
      Procedure.G.add_edge g (Procedure.Vert.End id) Procedure.Vert.Return)
    proc
```

katrinafyi (commented):
Thanks for this, it's good! There are some small readability comments which are hopefully easy to fix.

Forgive me for my ignorance because I haven't read Nick's thesis, but I also have some questions. 

I would really appreciate some module-level documentation about the purpose and idea of this transform. I think it's simple enough that I can guess the ideas, but it would be good to see it written out (at minimum, what is a CFA). Is there any reference you can link to?

Also, using `fold_blocks_topo_fwd` will amalgamate all of the edges from all blocks in the procedure, and it will not necessarily be in any sensible control-flow order (e.g., by dominators). Is this intentional and it won't cause any problems?

Edit: Ah, I guess it is sufficient to ensure that preds are all defined in the hashtbl before being accessed. However, the accumulation of `final_edge` is what is unexpected to me.

mira-alford:
Thanks for all the feedback!

I added an explanation and some more docs, let me know if it satisfies. Unfortunately there is not much to reference as its from an old draft snippet of the thesis that cut...

CFA is basically just a CFG by a different name too. Its unfortunate, i did consider just not using the term.

katrinafyi (approved):
Looks good, thanks for the updates!

I will note that it could be silently wrong if given a cyclic procedure, because `fold_blocks_topo_fwd` uses a weak topological order. But no big deal at the moment.

Edit: it is also quadratic because of repeatedly appending to a cons-list. something to keep in mind if you notice it's a bottleneck.

Issue #208, opened 2026-07-08
Fix the typing of generated functions


agle:
The memory encoding/boogie backend is incorrect currently as it gives the function definitions the type of their return value rather than the type of their function. The problem is the boogie backend to generate the correct (uncurried) declaration needs to identify the return type, which is a function/map type for some. Just returning the last type is wrong in this case as we can't distinguish `a -> b -> (c -> d)` from `a -> b -> c -> d`, we return `d` when we want `c -> d`.

We need to be able to write code essentially like below, assuming a mutable underlying implementation of unification through `unify`. Alternatively we could represent function types as uncurried but I'm wary of the impacts of having non-canonical forms of types floating around.

```ocaml
let ret_type = newtype () in
let new_fntype = curry  arg_types x in
let utype = unify new_fntype func_type in
ret_type
```

I have a partially complete implementation of hindley milner type inference, this aims to fix this but is a large change set. This issue covers just this function unification algorithm, which I will try to merge independently.

PR #216, opened 2026-07-09
Hindley-milnder type inference

agle:
This forms the skeleton of the hindley-milner type inference system. This comprises:

- A minimal hash-consed type expression representation in a union-find structure used for type inference (typeExpr.ml). 

This is essentially:

```ocaml
type 'a type_expr = Var of tvar | TypeConstr of 'a list * string
```
This should be sufficiently general to represent anything we want.

- Extraction from bincaml expressions to this representation
- Union-find based HM type unification/inference
- Elaboration pass over program

I expect the current inference and elaboration to not cover the whole program structure yet. 

- [x] I will write more on the design later to make it reviewable hopefully.
- [x] I will consider this mergable when we have some tests and a function for local expression type inference which we can use to solve the problem with the boogie backend: #208. I've had it floating around a while and want to get it in so I can start the migration without getting conflicts.~~

Similarly it doesn't do anything special to make sure the polymorphic ops get typed, identifying type errors, or let-generalisation. It doesn't do much yet as everything comes in typed already from the frontend; that will need a refactor to not enforce typing immediately.

We still generally use online type propagation from variables to get types, over the explicit `typ` field in each exprssion node; that will also need to be fixed in order to migrate to this system.

### DIFF
--- a/lib/lang/hm/elaboration.ml
+++ b/lib/lang/hm/elaboration.ml
@@ -1,5 +1,4 @@
 open Common
-open UnionFind
 open Abstract_expr
 
 (** Hindley-Milner type inference based on a union-find. *)
--- a/lib/lang/hm/elaboration.ml
+++ b/lib/lang/hm/elaboration.ml
@@ -1,0 +1,248 @@
+open Common
+open UnionFind
+open Abstract_expr
+
+(** Hindley-Milner type inference based on a union-find. *)
+
+module TypeInference (T : TypeExpr.TypeContext) = struct
+  include Inference.Make (T)
+
+  let retype_var univ ctx id =
+    lookup_var_typ ~no_constraint:true univ ctx id |> find |> to_basil
+    |> fun typ -> Var.copy ~typ id
+
+  let elaborate_phi univ ctx (p : Var.t Block.phi list) =
+    let open Block in
+    List.map
+      (fun { lhs; rhs } ->
+        let lhs = retype_var univ ctx lhs in
+        let rhs = List.map (fun (a, r) -> (a, retype_var univ ctx r)) rhs in
+        { lhs; rhs })
+      p
+
+  let ctx_to_string ctx =
+    TCtx.to_iter ctx
+    |> Iter.to_string (fun (a, b) ->
+        Printf.sprintf "%s %s" (V.to_string a) (scheme_to_string b))
+
+  let fresh_tvar ?(n = "a") () = fix @@ Var (gen.fresh ~name:n ())
+  let unfix i = match i with Expr.BasilExpr.E i -> i
+  let rec cata alg e = (unfix %> AbstractExpr.map (cata alg) %> alg) e
+
+  (** Extract type after full inference has run. *)
+  let elaborate_expr ~univ (hr : Lexing.position) e (c : scheme TCtx.t) =
+    let alg e =
+      let e =
+        match e with
+        | AbstractExpr.RVar { id; attrib; typ } ->
+            (* avoid looking up the id in context as we may be in a local
+            binding *)
+            let id = Var.copy id ~typ:(to_basil @@ find typ) in
+            AbstractExpr.RVar { id; attrib; typ }
+        | o -> o
+      in
+      let t = AbstractExpr.get_typ e |> find |> to_basil in
+      AbstractExpr.set_typ e t |> Expr.BasilExpr.fix
+    in
+    e |> TypingExpr.cata alg
+
+  let elaborate_stmt univ ctx stmt =
+    let retype_var v =
+      let typ =
+        lookup_var_typ ~no_constraint:true univ ctx v |> find |> to_basil
+      in
+      Var.copy ~typ v
+    in
+    Stmt.map
+      ~f_expr:(fun e -> elaborate_expr ~univ [%here] e ctx)
+      ~f_lvar:retype_var ~f_rvar:retype_var stmt
+
+  let elaborate_block p univ ctx (b : ('a, 'b) Block.t) =
+    Block.map ~phi:(elaborate_phi univ ctx) (elaborate_stmt univ ctx) b
+
+  let assume_proc_decl ctx ?(no_constraint = false) (p : Program.proc) =
+    let globs = Var.Decls.values (Procedure.local_decls p) in
+    let formals_in = Procedure.formal_in_params p |> StringMap.values in
+    let formals_out = Procedure.formal_out_params p |> StringMap.values in
+    let univ = V.proc_univ @@ Procedure.id p in
+    let ctx =
+      Iter.fold
+        (decl_var_typ ~no_constraint univ)
+        ctx
+        (Iter.append globs @@ Iter.append formals_in formals_out)
+    in
+    ctx
+
+  let infer_proc vc prog ctx ?(no_constraint = false) (p : Program.proc) =
+    let spec = Procedure.specification p in
+    let univ = V.proc_univ @@ Procedure.id p in
+    let ibool_list b =
+      List.map
+        (fun a ->
+          let a = infer_expr vc ~univ [%here] a ctx in
+          let _ = unify (fix bool_type) (getty a) in
+          a)
+        b
+    in
+    let spec : ('a, 'b) Procedure.proc_spec =
+      {
+        requires = ibool_list spec.requires;
+        ensures = ibool_list spec.ensures;
+        rely = ibool_list spec.rely;
+        guarantee = ibool_list spec.guarantee;
+        captures_globs =
+          List.map (infer_var global_universe ctx) spec.captures_globs;
+        modifies_globs =
+          List.map (infer_var global_universe ctx) spec.modifies_globs;
+      }
+    in
+
+    let new_spec ctx : (Var.t, Expr.BasilExpr.t) Procedure.proc_spec =
+      {
+        requires =
+          List.map (fun e -> elaborate_expr ~univ [%here] e ctx) spec.requires;
+        ensures =
+          List.map (fun e -> elaborate_expr ~univ [%here] e ctx) spec.ensures;
+        rely = List.map (fun e -> elaborate_expr ~univ [%here] e ctx) spec.rely;
+        guarantee =
+          List.map (fun e -> elaborate_expr ~univ [%here] e ctx) spec.guarantee;
+        captures_globs = spec.captures_globs |> List.map fst;
+        modifies_globs = spec.modifies_globs |> List.map fst;
+      }
+    in
+
+    let ctx = assume_proc_decl ctx ~no_constraint p in
+    let bvlocks =
+      Procedure.iter_blocks_topo_fwd p
+      |> Iter.map (fun (i, b) -> (i, infer_block vc prog univ ctx b))
+      |> Iter.persistent
+    in
+
+    let elaborate_proc ctx =
+      Procedure.set_specification p (new_spec ctx)
+      |> (fun p ->
+      bvlocks
+      |> Iter.map (fun (bid, b) -> (bid, elaborate_block [%here] univ ctx b))
+      |> Iter.fold (fun p (bid, b) -> Procedure.update_block p bid b) p)
+      |> Procedure.map_formal_in_params (StringMap.map (retype_var univ ctx))
+      |> Procedure.map_formal_out_params (StringMap.map (retype_var univ ctx))
+    in
+    Logs.debug (fun m -> m "%s" (ctx_to_string ctx));
+    (elaborate_proc, ctx)
+
+  (** Run type inference on a declaration, returning an updated typing scheme,
+      and elaboration function*)
+  let infer_decl visit_constraint prog scheme =
+    let open Program in
+    let infer_expr = infer_expr visit_constraint in
+    let infer_proc = infer_proc visit_constraint in
+    (* We have to be careful that inference is run immediately, not delayed until elaboration. *)
+    fun (decl_id, d) ->
+      match d with
+      | Type { binding; typ } ->
+          let ty = ty_of_basil typ in
+          let scheme = decl_type scheme binding ty in
+          let nty scheme = Type { binding; typ = to_basil (find ty) } in
+          (scheme, `Decl (decl_id, nty))
+      | Function { binding; definition; attrib } -> (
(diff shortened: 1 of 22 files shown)


lib/lang/hm/hm.ml
```diff
@@ -0,0 +1,91 @@
...
+    value-dependent bitvector widths, this will fail (emitting a type varable)
+    unless there are sufficient type annotations.
+
+    {3 Ad-hoc polymorpic operators}
+
+    Bincaml has some specific concerns, namely our operators are ad-hoc
+    polymorphic. For each op (e.g. bvadd)
+
+    A type-scheme in HM is a non-recursive universal quantifier over type
+    variables.
```
katrinafyi: unfinished sentence or needs colon or bullet point?

katrinafyi (commented):
My first thoughts are that I think the inference.ml should be split up. There's a few separate things happening inside there. Forgive any imprecision but my impression is that it's roughly...
- expression-level type unification
- intrinsic-level type scheme definition and instantiation (?)
- statement-level
- procedure-level including parameters
- big everything procedure inference
- concretisation back to Bincaml types

I won't direct you about details but you get the idea. It would also let you write module-level docs for each idea closer to the code. 

Generally, I would also like to see more.. restrained use of single-letter module names and local opens where possible.

lib/lang/hm/typeExpr.ml
```diff
@@ -0,0 +1,106 @@
...
+  let hash he = function
+    | Var v -> ID.hash v
+    | TypeConstr (l, t) -> Hash.pair (Hash.list he) Hash.string (l, t)
+end
+
+(** Create the union find and hash-consing structure for recording type
+    relations. This must be done each time we want to perform inference in an
+    independent environment, in order to construct a fresh hash-consing and
+    union-find state. *)
+module Make () = struct
```
katrinafyi: This seems really weird and risks being misused. Can you rename the module to `Make_fresh_hashcons` or something to make it different to normal Makes like Map.Make?

I also can't see if there is a fresh union-find state. Where is that, or is it just a fresh hash cons state?

agle: Union find uses refs so its implicitly fresh by creating a new hashcons

lib/lang/hm/inference.ml
```diff
@@ -0,0 +1,339 @@
...
+    | #Ops.BVOps.binary_pred ->
+        let a = fv () in
+        curry [ bvunk a; bvunk a ] bool_type
+    | #Ops.BVOps.unary_unif ->
+        let a = fv () in
+        curry [ bvunk a ] (bvunk a)
+    | `INTNEG -> curry [ int_type ] int_type
+    | #Ops.IntOps.binary_pred -> curry [ int_type; int_type ] bool_type
+    | #Ops.IntOps.const -> fix @@ int_type
+    | #Ops.IntOps.binary_unif -> curry [ int_type; int_type ] bool_type
```
katrinafyi: typo?


```suggestion
    | #Ops.IntOps.binary_unif -> curry [ int_type; int_type ] int_type
```

lib/lang/hm/solve_bv.ml
```diff
@@ -0,0 +1,72 @@
...
+    match a with
+    | Add { a; b; equ } -> (
+        match List.map is_nat_val_type [ find a; find b; find equ ] with
+        | [ Some _; Some _; Some _ ] -> Some equ
+        | [ Some a; Some b; None ] ->
+            Some (unify ~pos:[%here] (fix @@ nat_val_type (a + b)) equ)
+        | [ Some a; None; Some b ] ->
+            Some (unify ~pos:[%here] (fix @@ nat_val_type (b - a)) equ)
+        | [ None; Some a; Some b ] ->
+            Some (unify ~pos:[%here] (fix @@ nat_val_type (b - a)) equ)
```
katrinafyi: in the last two cases, bug? `a + b == equ` but this equation says `b - a == equ` ?

lib/lang/hm/solve_bv.ml
```diff
@@ -0,0 +1,72 @@
...
+        | _ -> None)
+    | AddConst { a; b; equ } -> (
+        match (is_nat_val_type (find a), is_nat_val_type (find equ)) with
+        | Some i, None ->
+            Some (unify ~pos:[%here] (fix @@ nat_val_type (i + b)) equ)
+        | Some i, Some equ_c ->
+            Some (unify ~pos:[%here] (fix @@ nat_val_type (i + b)) equ)
```
katrinafyi: question: why can this not just use equ_c?

lib/lang/hm/unification.ml
```diff
@@ -0,0 +1,124 @@
...
+    let b = find b in
+    raise
+      (TypeErr
+         ("recursive: tvar " ^ ID.to_string a ^ " occurs in " ^ type_to_string b))
+
+  (** Check for type recursion: recursion is failure. *)
+  let occurs_in a b =
+    let check = function
+      | Var t -> equal_tvar t a
+      | TypeConstr (ls, _) -> List.fold_left ( || ) false ls
```
katrinafyi:
```suggestion
      | TypeConstr (ls, _) -> List.for_all Fun.id ls
```

lib/lang/hm/unification.ml
```diff
@@ -0,0 +1,124 @@
...
+    Logs.debug (fun m ->
+        m "%s"
+          ("unify: " ^ plpos pos ^ " "
+          ^ type_to_string (find t)
+          ^ " with "
+          ^ type_to_string (find t')));
+    match (map_expr unfix @@ Typ.unfix t, map_expr unfix @@ Typ.unfix t') with
+    | TypeConstr ([], "nothing"), _ -> t
+    | _, TypeConstr ([], "nothing") -> t'
+    | Var x, Var y -> union t t'
```
katrinafyi: does this lead to infinite loop? i can't see why this wouldn't just keep hitting this case and recursing.

lib/lang/hm/unification.ml
```diff
@@ -0,0 +1,124 @@
...
+    | Var x, _ when occurs_in x t' -> recursion_error x t'
+    | Var _, TypeConstr _ -> merge (fun a b -> b) t t'
+    | _, Var x -> unify ~pos:[%here] t' t
+    | TypeConstr ([ a ], "const"), TypeConstr ([ b ], "const") ->
+        let x = unify ~pos:[%here] (fix a) (fix b) in
+        fix @@ TypeConstr ([ x ], "const")
+    | TypeConstr ([ a ], "shared"), TypeConstr ([ b ], "shared") ->
+        let x = unify ~pos:[%here] (fix a) (fix b) in
+        fix @@ TypeConstr ([ x ], "const")
```
katrinafyi: bug/typo?

```suggestion
        fix @@ TypeConstr ([ x ], "shared")
```

katrinafyi (commented):
important comments/questions

lib/lang/hm/solve_bv.ml
agle: This is just to check that the constraint holds

lib/lang/hm/unification.ml
agle: Ill put module path in

lib/lang/hm/hm.ml
```diff
@@ -0,0 +1,198 @@
...
+  let _ = T.solve_constraints ~max_iters:100 !constraints in
+  let e = T.elaborate_expr ~univ [%here] i ctx in
+  e
+
+(** Algebra for returning the annotated type (for use with functions like
+    fold_with_type)*)
+let elaborated_type_alg (e : Types.t Expr.BasilExpr.abstract_expr) =
+  Expr.AbstractExpr.get_typ e
+
+(* Partially apply args list to function type funtype and return resulting type *)
```
katrinafyi:
```suggestion
(** Partially apply args list to function type funtype and return resulting type *)
```

lib/lang/hm/hm_types.ml
```diff
@@ -0,0 +1,121 @@
+(** Conversion from bincaml types to and from HM type infernece environment *)
```
katrinafyi:
```suggestion
(** Conversion from bincaml types to and from HM type inference environment *)
```

lib/lang/hm/solve_bv.ml
```diff
@@ -0,0 +1,72 @@
...
+    | Add { a; b; equ } ->
+        Printf.sprintf "%s + %s = %s" (type_to_string a) (type_to_string b)
+          (type_to_string equ)
+    | AddConst { a; b; equ } ->
+        Printf.sprintf "%s + (const %d) = %s" (type_to_string a) b
+          (type_to_string equ)
+
+  (** Deduce the width if two values of the constraint have inferred widths *)
+  let unify_bv_constraint a : Typ.t option =
+    match a with
```
katrinafyi:
```suggestion
  let unify_bv_constraint constr : Typ.t option =
    match constr with
```

lib/lang/hm/unification.ml
```diff
@@ -0,0 +1,124 @@
...
+        let v = merge (fun _ c -> c) (fix vr) (fix cst) in
+        fix @@ TypeConstr ([ v ], "bv")
+    | TypeConstr (ars, n), TypeConstr (ars', n') ->
+        let args =
+          List.combine ars ars'
+          |> List.map (fun (a, b) -> unify ~pos:[%here] (fix a) (fix b))
+        in
+        merge (fun a b -> TypeConstr (args, n)) t t'
+
+  (* instantiate typescheme for a single type-annotated variable *)
```
katrinafyi:
```suggestion
  (** instantiate typescheme for a single type-annotated variable *)
```

lib/lang/hm/unification.ml
```diff
@@ -0,0 +1,124 @@
...
+  let inst_annot_v ?(no_constraint = false) v =
+    let ty =
+      match Var.typ v with
+      | _ when no_constraint -> fix @@ Var (gen.fresh ~name:(Var.name v) ())
+      | Nothing -> fix @@ Var (gen.fresh ~name:(Var.name v) ())
+      | o -> ty_of_basil o
+    in
+    ty
+
+  (* lookup var and add unify with its type annotation *)
```
katrinafyi:
```suggestion
  (** lookup var and add unify with its type annotation *)
```

lib/lang/hm/unification.ml
```diff
@@ -0,0 +1,124 @@
...
+    let a =
+      let v = V.of_var univ v in
+      TCtx.find_opt v c |> function
+      | Some v -> v
+      | None -> failwith ("var not found: " ^ V.to_string v)
+    in
+    let tt = match a with Forall (_, ty) -> union ty vt in
+    tt
+
+  (* declare type with name in type scheme *)
```
katrinafyi:
```suggestion
  (** declare type with name in type scheme *)
```

lib/lang/hm/unification.ml
```diff
@@ -0,0 +1,124 @@
...
+  let decl_type ctx name vt =
+    let tvar = V.create types_universe name in
+    TCtx.update tvar
+      (function
+        | Some (Forall ([], t)) -> Some (Forall ([], unify vt t))
+        | None -> Some (Forall ([], vt))
+        | _ -> failwith "unk")
+      ctx
+
+  (* declare var with type in type scheme *)
```
katrinafyi:
```suggestion
  (** declare var with type in type scheme *)
```

lib/lang/hm/hm.ml
katrinafyi: ^ i can't see any changes near this comment and the punctuation is still wrong?

lib/lang/hm/elaboration.ml
```diff
@@ -0,0 +1,248 @@
...
+                        (elaborate_expr ~univ:global_universe [%here] b scheme);
+                  }
+              in
+              (scheme, `Decl (decl_id, nb))
+          | Uninterpreted ->
+              let nu s =
+                Function
+                  { binding = binding s; attrib; definition = Uninterpreted }
+              in
+              (scheme, `Decl (decl_id, nu))
```
katrinafyi: naming. what is `nu`? it would be good to rename `nu` and all the other `nX` functions in this scope.

lib/lang/hm/hm.ml
```diff
@@ -0,0 +1,198 @@
+(** Simple hindley-milner type inference *)
+
```
katrinafyi, line 2: More introduction documentation please. For example, add links to https://bernsteinbear.com/blog/type-inference/ and https://en.wikipedia.org/wiki/Hindley%E2%80%93Milner_type_system#Algorithm_J

lib/lang/hm/typeExpr.ml
```diff
@@ -0,0 +1,109 @@
...
+
+module TCtx = Map.Make (V)
+(** The type scheme / typing context : to store a map from scoped type variables
+    to types. *)
+
+module ATyp = struct
+  (** Open recursive type expression, either a variable or type constructor. *)
+
+  type 'a expr = Var of tvar | TypeConstr of 'a list * string
+  [@@deriving eq, ord, show, map, fold]
```
katrinafyi, line 41: I understand the appeal of the simplicity of the string-named type constructors, but it is a significant risk for typos and unmatched cases. There is also a more terrifying risk that a user-defined type will collide with a built-in special type.

Aside from user-defined records, all type constructor names are all static so it seems to me like an enum would be both easy and improve safety. But this is too late so it doesn't have to change now.

katrinafyi (commented):
I don't think it's possible to review this PR properly, so I am really just making sure that it's documented enough and organised enough.

This will be the last major round of comments.

lib/lang/hm/typeExpr.ml
agle: Since the parser won't let you define a  type named `->` this is more a quality of type errors issue (which will certainly occur if you shadow a builtin type)  than a correctness issue. It makes the code slightly more error prone but everything's behind smart constructors and the only thing that matches on constructor names is unification and conversion between bincaml types. 

I somewhat prefer the formal simplicity here, it allows the data-structure to be reused for a different type system. In a sense an enum would be less safe as lets you represent bool as `Bool` and `Custom "bool"`---you would have to explicitly parse/validate that you don't shadow a built-in type with a custom one.

PR #217, opened 2026-07-10
Flat map statements into blocks, and use in Aslp

katrinafyi:
This adds a function for mapping a single statement into one or more blocks (or statements), while fixing up all the control flow.
```ocaml
type ('v, 'e) mapped_stmt =
  [ `Stmts of (Var.t, Var.t, BasilExpr.t) Stmt.t list
  | `Blocks of (Var.t, BasilExpr.t) Block.t list
  | `Graph of ID.t * ID.t * ('v, 'e) t ]

let general_flat_map_stmts
    ~(f : proc:_ t -> _ Stmt.t -> _ mapped_stmt)
    base_bid
    proc
```
The function `f` should return multiple blocks by returning a `(first_id, last_id, proc')` where `first` and `last` are new blocks which `f` inserts into the updated `proc'`. Alternatively, it may also return a `stmt_list`. In this case, it is assumed that the procedure doesn't change.

The implementation is a bit big but hopefully it's normal. As well as being useful for Aslp, I'm sure this will be useful in other cases too. For example, maybe for inlining procedure calls.

This PR also changes the Aslp frontend to make use of the new `flat_map_stmts`. This leads to much simpler code. The Aslp transform no longer needs to split the statement list into before+intrin+after, since we can just handle one statement at a time. Impressively, there are very minimal changes to the expect test output.

It also fixes the quadratic runtime of the current code in main. There's also a expect test to verify this. If you want, this can be deleted since timing-based tests can easily become flaky.

I have also left some TODO comments in the diff and want some advice on those. I'm also not sure about the name of the function, there's already a few functions with similar names..


agle:
> The function f should return multiple blocks by returning a Left (first_id, last_id, proc') where first and last are new blocks which f inserts into the updated proc'.

Does this strictly insert two blocks from proc' into proc? Or does it take all blocks between these assuming a diamond? What happens if the caller makes unrelated changes to the procedure in `~f`?


```ocaml
Var.t, Var.t, Program.e) Stmt.t ->
    ( ID.t * ID.t * ('a, 'b) Procedure.t,
      (Var.t, Var.t, Program.e) Stmt.t list )
    Either.t
```

This should be a defined type or polymorphic variant with names for the cases. 

What is the `base_bid`, last `ID.t` argument?

katrinafyi:
I think I've done the suggestions now :) The functions are moved around, renamed, and better documented.

lib/lang/procedure.ml
```diff
@@ -649,6 +666,127 @@ let flat_map_stmts_topo_rev ?visit rewriter p =
           update_block p bid (Block.flat_map ~rev:true ~phi:Fun.id rewriter b))
     p blocks
 
+type ('v, 'e) mapped_stmt =
+  [ `Stmts of (Var.t, Var.t, BasilExpr.t) Stmt.t list
+    (** Zero or more straight-line statements. *)
+  | `Blocks of (Var.t, BasilExpr.t) Block.t list
+    (** Zero or more straight-line blocks. *)
+  | `Graph of ID.t * ID.t * ('v, 'e) t
```
agle, line 674: If declaring the type it may as well be a regular variant as it provides better type inference and errors.

lib/lang/procedure.ml
```diff
@@ -649,6 +666,127 @@ let flat_map_stmts_topo_rev ?visit rewriter p =
...
+
+        Generally, the procedure should not be modified in any other way. *) ]
+(** Either a list of zero or more statements, {i or} a first/last pair of
+    block-level control-flow. This is used as the expected return type of
+    mapping one statement in {!general_flat_map_stmts}'s [~f] parameter. *)
+
+(** [general_flat_map_stmts ~f bid proc] maps each statement in the given block
+    ID through [f]. For each statement, [f] may return either zero or more
+    "bare" statements, {i or} a first/last pair of new block-level control-flow.
+    See {!type-mapped_stmt} for details about [f]'s return type.
```
agle: Or list straight-line list of blocks.

lib/lang/procedure.ml
```diff
@@ -649,6 +666,127 @@ let flat_map_stmts_topo_rev ?visit rewriter p =
...
+    find_block proc base_bid |> Block.stmts_iter |> Iter.to_list
+    |> List.fold_flat_map
+         (fun (proc, bid) stmt ->
+           match f ~proc stmt with
+           | `Blocks [] | `Stmts [] -> ((proc, bid), [])
+           | `Blocks bs ->
+               let proc, bids =
+                 List.fold_map
+                   (fun proc b ->
+                     let[@warning "+9"] { Block.attrib; stmts; phis } = b in
```
agle: What is this for? could use warning name

`9 [missing-record-field-pattern] Missing fields in a record pattern.`

lib/lang/procedure.ml
```diff
@@ -649,6 +666,127 @@ let flat_map_stmts_topo_rev ?visit rewriter p =
...
+    [first] / [last] is the first / last block of the combined map output.
+
+    The returned [first] and [last] may be the same. One or both of
+    [first]/[last] may be the same as the original block. In particular, any
+    bare statements returned by [f] for an initial segment of the block's
+    statements will be retained in the original block.
+
+    Additionally, redirects the original block's incoming/outgoing edges to the
+    first / last block of the mapped output. *)
+let general_flat_map_stmts ~(f : proc:_ t -> _ Stmt.t -> _ mapped_stmt) base_bid
```
agle: > redirects the original block's incoming/outgoing edges to the
    first / last block of the mapped output

This seems subtle—trivial for `Stmts`, obvious for `Blocks`, but the `Graph` case could do odd things like remove the successors of the original block, or add additional successors. I think it should be made clear that the `Graph` case need not produce an additional subgraph that is connected to the original? Its just a new subgraph detached in the procedure (but in the same forest) that gets linked up by this function?

> One or both of [first]/[last] may be the same as the original block.

Not sure if this case is more confusing to allow; might want to force to return blocks or stmts in that case.

Could a version of the API generate fresh begin + end blocks of the diamond and pass them to a closure returned by the `Graph` case to inject the diamond between? That way you would be appending / prepending to a block and have it be automatically bookended; it could check the succ/pred sets of the end points aren't modified. Maybe it could simplify the other special cases where it can concat to a prev block as well.

Sorry I say all this because its not obvious how I would use this immediately; e.g. what assumption do you make of the state of the graph within `f` when processing an arbitrary stmt of the program in order to modify the graph? And what do you have to guarantee about the graph you return?

Maybe the subgraph refactor would be the easiest way to clean that up though.

lib/lang/procedure.ml
```diff
@@ -649,6 +666,127 @@ let flat_map_stmts_topo_rev ?visit rewriter p =
...
+    proc =
+  let existing_bids =
+    lazy (iter_blocks proc |> Iter.map fst |> Iter.to_set (module IDSet))
```
agle: Why not IDSet.of_iter?

lib/lang/procedure.ml
```diff
@@ -649,6 +666,127 @@ let flat_map_stmts_topo_rev ?visit rewriter p =
...
+                    block IDs"
+           | `Stmts s ->
+               let name = ID.name base_bid and stmts = [] in
+               let proc, bid =
+                 match bid with
+                 | None -> fresh_block proc ~name ~stmts ()
+                 | Some bid -> (proc, bid)
+               in
+               ((proc, Some bid), [ Either.Right (bid, Iter.of_list s) ]))
+         (proc, Some base_bid)
```
agle: Probably factor out the body of this fold function.

lib/lang/procedure.ml
```diff
@@ -649,6 +666,127 @@ let flat_map_stmts_topo_rev ?visit rewriter p =
...
+           | `Blocks bs ->
+               let proc, bids =
+                 List.fold_map
+                   (fun proc b ->
+                     let[@warning "+9"] { Block.attrib; stmts; phis } = b in
+                     fresh_block proc ~attrib ~phis
+                       ~stmts:(CCVector.to_list stmts) ())
+                   proc bs
+               in
+               ((proc, None), List.map (fun x -> Either.Left (x, x)) bids)
```
agle: What are `Left` and `Right` representing in this result list?

lib/lang/procedure.ml
```diff
@@ -649,6 +666,127 @@ let flat_map_stmts_topo_rev ?visit rewriter p =
...
+      List.last_opt block_id_pairs |> Option.map_or snd ~default:base_bid )
+  in
+  let proc =
+    proc
+    |> transplant_outgoing_edges ~from:base_bid ~to_:last
+    |>
+    if not ID.(equal first base_bid) then
+      add_goto ~from:base_bid ~targets:[ first ]
+    else Fun.id
+  in
```
agle, line 772: Might help clarify the API if you also require that `f : Graph` case does not modify the control flow of the original block.

lib/lang/program.ml
```diff
@@ -284,6 +284,16 @@ let flat_map_decls f p =
...
 
+(** Iterates over global variables in the given program, including both read and
+    assigned variables. Order is unspecified and may have duplicates. *)
+let referenced_vars_of_prog =
+  procs
+  %> Iter.flat_map
+       (snd %> Procedure.iter_blocks
+       %> Iter.flat_map (fun (_, b) ->
+           Iter.append (Block.read_vars_iter b) (Block.assigned_vars_iter b)))
+  %> Iter.filter Var.is_global
```
agle, line 295: It should be a requirement that this should always be less than  [`Program.global_vars`](https://agle.github.io/bincaml/bincaml/bincaml.ast/Lang/Program/index.html#val-global_vars). It should be covered by checks.ml.

lib/lang/block.ml
```diff
@@ -106,11 +106,14 @@ let map ~phi f (b : ('v, 'e) t) : ('vv, 'ee) t =
   { stmts = Vector.map f b.stmts; phis = phi b.phis; attrib = b.attrib }
 
 (** Modify stmt list by creating a mutable copy of the underlying vector *)
-let fmap_stmts_copy (f : (('a, 'a, 'b) Stmt.t, 'c) Vector.t -> unit) b =
+let fmap_stmts_copy (f : (('a, 'a, 'b) Stmt.t, 'mut) Vector.t -> unit) b =
```
agle: Should `'mut` be the specifc mutable type defined in Vector rather than generic?

agle (approved):

lib/lang/procedure.ml
katrinafyi: Variants are kinda annoying to use because you have to qualify them with the module. Also, a variant like `Procedure.Blocks` is rather confusing I think, so it would need a different name anyway. And it's really just a special type used in this one function that is only declared separately so I can ocamldoc it.

Unless you feel strongly, I will leave it.

lib/lang/procedure.ml
katrinafyi: ocamlformat does this
```ocaml
                     let[@warning "+missing-record-field-pattern"] {
                                                                 Block.attrib;
                                                                 stmts;
                                                                 phis;
                                                               } =
                       b
```

i'll see what i can do

lib/lang/procedure.ml
katrinafyi: That's cool but I'm probably not going to do that (have the `flat_map` implementation provide bookend blocks). I don't think it's a qualitative improvement over the current approach if we're still relying on Procedure to store the block-level control-flow.

> This seems subtle ...

> Sorry I say all this because its not obvious how I would use this immediately

`f`'s purpose is to map one statement. So, `f` should make no assumptions at all on the state of the graph.

Are your questions adequately answered by the doc comment on `` `Graph``? If not, how can it be improved? I think it answers all of your questions, just not in the same graph theory words.

lib/lang/program.ml
katrinafyi: Program.global_vars filters the decls list, but here we need to get referenced variables which may or may not have been declared.

lib/lang/block.ml
katrinafyi: no. in CCVector, rw and ro are two totally distinct types.

lib/lang/procedure.ml
katrinafyi: did this

lib/lang/procedure.ml
katrinafyi: moved into a local function and added a comment above that local function

lib/lang/procedure.ml
katrinafyi: I didn't know it 🤯

lib/lang/procedure.ml
katrinafyi: added the warning annotation onto an outer let

lib/lang/block.ml
agle: I know that, I thought it was being passed a mutable vector specifically, but `Vector.copy` is shallow and just returns the same vector if its immutable.

PR #218, opened 2026-07-15
Aslp transform + fixes

agle:
Transform that runs the aslp lifter + inlining of results into program.

Some fixes to the stmt/expr translation based on checks in `test/cram/gtirb.t`


lib/lang/spec_modifies.ml
```diff
@@ -76,14 +76,17 @@ let set_modsets ?(add_only = false) prog =
         |> Iter.flat_map Expr.BasilExpr.free_vars_iter
         |> Iter.filter Var.is_global |> VarSet.of_iter
       in
+      let name_compare = fun a b -> String.compare (Var.name a) (Var.name b) in
       let captures_globs =
-        List.filter (not % Var.is_constant)
+        List.sort name_compare
+        @@ List.filter (not % Var.is_constant)
```
katrinafyi: while you're here, using pipe is a bit more readable than long chains of `@@`.

lib/transforms/aslp/aslp_state.ml
```diff
@@ -102,7 +102,7 @@ let empty_aslp_ids () =
     This will ensure that ASLp's local variable and block names do not clash
     with existing names. *)
 let aslp_ids_from_generators ~local_ids =
-  let local_id = ID.fresh ~name:"var" local_ids %> ID.name in
+  let local_id = ID.fresh ~name:"lv" local_ids %> ID.name in
```
katrinafyi: lv could be confused with "local var"? could try "local"?

lib/invariants.ml
```diff
@@ -16,13 +16,15 @@ type t =
   | Params
   | LambdaLift
   | MemoryEncoding
+  | GTIRB_ARM
```
katrinafyi: should it be camelcase for consistency idk.
```suggestion
  | GtirbArm
```

lib/lang/spec_modifies.ml
```diff
@@ -76,14 +76,17 @@ let set_modsets ?(add_only = false) prog =
         |> Iter.flat_map Expr.BasilExpr.free_vars_iter
         |> Iter.filter Var.is_global |> VarSet.of_iter
       in
+      let name_compare = fun a b -> String.compare (Var.name a) (Var.name b) in
```
katrinafyi:
```suggestion
      let name_compare = CCOrd.map Var.name String.compare in
```
https://c-cube.github.io/ocaml-containers/containers/CCOrd/index.html#val-map

lib/transforms/aslp/aslp.ml
```diff
@@ -73,11 +73,34 @@ let stmt_of_aarch64_intrin ?error :
...
-  | _ -> failwith "aarch64_mem_of_prog: no $mem found"
+  | Some (Variable { binding }) ->
+      if
+        not
+          (Types.equal (Var.typ binding)
+             Types.(Map (Bitvector 64, Bitvector 8)))
+      then
+        Logs.warn (fun m ->
+            m
+              "Declared with unexpected type; lifter may not produce \
```
katrinafyi:
```suggestion
              "Memory declared with unexpected type; lifter may not produce \
```

lib/transforms/aslp/aslp.ml
```diff
@@ -73,11 +73,34 @@ let stmt_of_aarch64_intrin ?error :
...
+               well-typed or correct code.");
+      (prog, binding)
+  | None ->
+      let mem =
+        Var.create "$mem" ~scope:Var.GlobalVarShared
+          Types.(Map (Bitvector 64, Bitvector 8))
+      in
```
katrinafyi: could factor out the name and type because it's repeated here and in the warning check above.

lib/transforms/aslp/bincaml_ibi_make.ml
```diff
@@ -375,20 +381,30 @@ struct
   let f_gen_slt_bits : bigint -> expr -> expr -> expr =
    fun _ a b -> Expr.BasilExpr.binexp ~op:`BVSLT a b
 
+  let make_widths_equal a b =
```
katrinafyi: these arguments are meaningful and not interchangeable, so generic names like `a` and `b` aren't so good. maybe make the width reference value into a named parameter?

lib/script.ml
```diff
@@ -315,6 +315,7 @@ let load_gtirb st fname =
       let p =
         match p with
         | Ok e ->
+            let e = Program.set_attrib Invariants.(to_attrib [ GTIRB_ARM ]) e in
```
katrinafyi: the invariant should be set by `Gtirb_frontend.load_gtirb_prog`, not here after the fact.

lib/transforms/aslp/bincaml_ibi_make.ml
```diff
@@ -297,25 +297,31 @@ struct
...
+  (** [let lhs = f_gen_Mem_read size address size acctype value] emits a read
+      instruction reading a bitvector [size] (bytes) from the memory variable,
+      assigning it to variable [lhs]. Returns [lhs] as an expression). *)
   let f_gen_Mem_read : bigint -> expr -> expr -> expr -> expr =
    fun size addr _ _acctype ->
-    let addr = Stmt.Addr { addr; size = Z.to_int size; endian = `Little }
+    let size = 8 * Z.to_int size in
+    let addr = Stmt.Addr { addr; size; endian = `Little }
     and mem = S.bincaml_memory_var () in
     let name = !bincaml_lifter_state.generator.local_id () in
```
katrinafyi, line 321: i had a question but i answered it myself. i'd like to record it in a comment.
```suggestion
    (* NOTE: avoids `bincaml_local_var` so these variables do not clash with ASLp-declared variables. *)
    let name = !bincaml_lifter_state.generator.local_id () in
```

lib/transforms/aslp/bincaml_ibi_make.ml
```diff
@@ -375,20 +381,30 @@ struct
   let f_gen_slt_bits : bigint -> expr -> expr -> expr =
    fun _ a b -> Expr.BasilExpr.binexp ~op:`BVSLT a b
 
+  let make_widths_equal a b =
+    let width a =
+      Expr.BasilExpr.type_of a |> Types.bit_width
+      |> Option.get_exn_or "expected bv argument to operator"
+    in
```
katrinafyi, line 389: Consider moving this helper function (or, indeed the whole `make_widths_equal` function) into expr.ml.

katrinafyi (commented):
Thanks for the IBI fixes. 

The most important comments are about position of `set_attrib` for invariants, and for naming around `make_widths_equal`.

lib/transforms/aslp/aslp_state.ml
agle: Why is interpreting it as "local var" a confusion, it is a local var?

lib/script.ml
agle: It has to be defined here because `Gtirb` doesn't link `Passes.Invariants` -- would be a cycle. You could redefine `to_attrib GTIRB_ARM` in `Gtirb` but that seems worse.

lib/transforms/aslp/aslp_state.ml
katrinafyi: sorry - i was thinking of left var. but regardless, i think local is better and clearer

lib/transforms/aslp/bincaml_ibi_make.ml
```diff
@@ -375,20 +382,34 @@ struct
...
 
   let f_gen_lsl_bits : bigint -> bigint -> expr -> expr -> expr =
-   fun _ _ a b -> Expr.BasilExpr.binexp ~op:`BVSHL a b
+   fun _ _ a b ->
+    Expr.BasilExpr.binexp ~op:`BVSHL a (make_shift_extended_to_value_width a b)
 
   let f_gen_asr_bits : bigint -> bigint -> expr -> expr -> expr =
-   fun _ _ a b -> Expr.BasilExpr.binexp ~op:`BVASHR a b
+   fun _ _ a b ->
+    Expr.BasilExpr.binexp ~op:`BVASHR a (make_shift_extended_to_value_width a b)
```
katrinafyi: Changes it to `extend_to_match ~width_of x`.

I was mostly worried about confusing the two arguments, not the naming of the function itself. There's no need to have such a clunky name.

```suggestion
  let extend_to_match ~width_of x =
    let width a =
      Expr.BasilExpr.type_of a |> Types.bit_width
      |> Option.get_exn_or "expected bv argument to operator"
    in
    let wd_val = width x and wd_target = width width_of in
    if wd_val = wd_target then x
    else if wd_target > wd_val then
      Expr.BasilExpr.zero_extend ~n_prefix_bits:(wd_target - wd_val) x
    else failwith "expected argument to have equal or smaller width"

  let f_gen_mul_bits : bigint -> expr -> expr -> expr =
   fun _ a b -> Expr.BasilExpr.applyintrin ~op:`BVMUL [ a; b ]

  let f_gen_append_bits : bigint -> bigint -> expr -> expr -> expr =
   fun _ _ a b -> Expr.BasilExpr.applyintrin ~op:`BVConcat [ a; b ]

  let f_gen_lsr_bits : bigint -> bigint -> expr -> expr -> expr =
   fun _ _ a b ->
    Expr.BasilExpr.binexp ~op:`BVLSHR a (extend_to_match ~width_of:a b)

  let f_gen_lsl_bits : bigint -> bigint -> expr -> expr -> expr =
   fun _ _ a b ->
    Expr.BasilExpr.binexp ~op:`BVSHL a (extend_to_match ~width_of:a b)

  let f_gen_asr_bits : bigint -> bigint -> expr -> expr -> expr =
   fun _ _ a b ->
    Expr.BasilExpr.binexp ~op:`BVASHR a (extend_to_match ~width_of:a b)
```

lib/transforms/aslp/bincaml_ibi_make.ml
agle: No, its too trivial.

lib/script.ml
katrinafyi: Okay, let's leave it for now.

It's true that it doesn't quite fit into the existing system. In other cases, the pass manager manages the invariants but we can't do that here because it's the initial loading of the program.

To be honest, the ASLp transform doesn't really need an intrinsic. It should be an no-op if run on an IR with no aarch64 intrinsic calls.

lib/transforms/aslp/aslp_state.ml
katrinafyi: Thanks.

lib/lang/spec_modifies.ml
```diff
@@ -76,14 +76,17 @@ let set_modsets ?(add_only = false) prog =
...
       let captures_globs =
-        List.filter (not % Var.is_constant)
-        @@ VarSet.elements @@ VarSet.union vs
-        @@ VarSet.union exist_captures
-        @@ VarSet.union read written
+        VarSet.union read written
+        |> VarSet.union exist_captures
+        |> VarSet.union vs |> VarSet.elements
+        |> List.filter (not % Var.is_constant)
+        |> List.sort name_compare
```
katrinafyi, line 85: The piping made me notice that it's making lots of intermediate varsets which aren't needed. Sorry for the repeated changes. 

```suggestion
        [ read; written; exist_captures; vs ]
        |> List.map VarSet.to_iter |> Iter.append_l
        |> Iter.filter (not % Var.is_constant)
        |> Iter.to_list |> List.sort name_compare
```

lib/transforms/aslp/aslp.ml
```diff
@@ -170,13 +188,13 @@ let add_aarch64_global_declarations ?(add_all = false) prog =
     procedures within the given program.
 
     Also inserts global variable declarations for the architectural variables,
-    if not already present. *)
+    if not already present. Assumes it is running immediately after gtirb. *)
```
katrinafyi: As said in the other comment, I had planned that it should be an no-op if run on a program with no aarch64 intrinsic calls, so this assumption shouldn't be needed.

Is there a change in this PR which needs this assumption? If not, just remove the comment.

katrinafyi (approved):
LGTM modulo the make_shift_extended_to_value_width naming. I would still like answers to some other comments in this most recent review. Thanks.

lib/lang/spec_modifies.ml
agle: It unions all the sets to ensure each variable is only output once. Would you prefer union the sets in a fold, or do `List.sort_uniq ~cmp:Var.compare |> List.sort name_compare` or `List.sort Var.compare` which is sensitive to construction order or just `List.sort_uniq ~cmp:name_compare` which might more confusingly hide errors due to equally named differently-typed variables?

lib/transforms/aslp/aslp.ml
agle: Its probably not needed in that case but helpful to tell the user the expected running order (as most of the time they probably don't _want_ a noop).

Adding the mem if not exists will happen regardless on if it lifts any statements

(cant reply to the other comment)

lib/lang/spec_modifies.ml
katrinafyi: It's an interesting edge case. There are changes that could make it work but let's just abandon this thread and leave it.

lib/transforms/aslp/aslp.ml
katrinafyi: If it's about the suggested running order, then you can say that rather than enshrining it as a strict assumption. I dont want later changes to the transform code to require a pass order, unless it's absolutely necessary.

I acknowledge the memory adding will always happen once.

katrinafyi: To reply to the other comment, you have to click the timestamp and go to the original thread.

katrinafyi: If you don't want to change this right now, that is okay.

katrinafyi (approved):
Feel free to merge.

agle:
I still want to relax the expected time limit more on the linear cfg flat map test because its still being flaky and I think we have a couple orders of magnitude wiggle room between the current version and previous quadratic one.

katrinafyi:
Of course. Tbh I'd just delete the test if it's too troublesome. It was mostly for  TDD and is not at risk of regressions.

One of the problems with the timing test is that you need bigger input sizes to make the differences more apparent, and that just increases the test runtime for a rather trivial test.

agle:
Ok I'll delete it, I agree its not at risk of regressions. I did already increase the problem size to make the gap wider but the CI runners are probably too inconsistent for it to be reliable.

PR #219, opened 2026-07-21
SSIfy

NambatCat:
### Implementation of SSI algorithm:
The file ssify.ml implements the algorithm for converting a program to SSI form. This algorithm is based off Chapter 13.2 in the book “SSA-based Compiler Design”, which describes a 3-step algorithm as follows:

**Step 1: Splitting**

- The split algorithm traverses the CFG of the procedure and inserts phi nodes (at join nodes), sigma nodes (at branch nodes) and parallel copies appropriately.

- In our implementation, instead of explicitly defining sigma nodes, we use phi nodes to represent sigma nodes – refer to section 13.2.5 - “Implementing σ-Functions” in “SSA-based Compiler Design”.

**Step 2: Renaming**

- The rename algorithm takes the new CFG produced by the split function and renames variables to maintain the single static assignment property – this ensures that each variable is only assigned to once.

**Step 3: Clean**

- The clean algorithm takes the previously transformed CFG from the rename function and removes any dead and unreachable code, ensuring all redundant code is removed in the final output

This algorithm can be used to convert a program not in SSI form to SSI form, and to reconstruct a program that was previously in SSI form but is no longer in this form. The code has been thoroughly tested using various tests (loops, multiple dependences, reconstruction) to ensure edge cases are accounted for. The tests within the file can be run using `dune runtest`. Additional tests located in `test2.s` and `passes.ml` can be run using `dune exec bincaml script lib/test2.s`

Ideally, this algorithm should only be run on a program where many variables have broken SSI form, using `SSIfy.ssify_prog prog`, or on a specific variable located in the program that is known to be out of SSI form, using `SSIfy.ssify_prog_var_name var_name prog`

Todo:
- [x]  Run a flame test to determine which parts are the least efficient
- [x]  Make the code less imperative
- [ ]  Update the set of non-actual instructions all at once at the end of rename, instead of having nested maps
- [ ]  Improve the time performance of ssify, in particular rename and clean.

Known issues:
- [x]  The second while loop in clean that operates on inst.uses causes an infinite loop when left as-is, due to formal-in and out paramters. It is prevented by intersecting inst.uses with web, but this may cause deviations from the book's expected output.
- [x]  Multiple phi nodes for a single variable are being created when ssify_prog is run rather than ssify on a single variable, which causes rename and clean to take a significant amount of time editing and removing instructions that should not be created in the first place
- [ ]  Procedures with multiple return nodes incorrectly assign to a new formal-out variable

Contributors:
Namson Phan
Amna Arshad


lib/transforms/ssify.ml
```diff
@@ -0,0 +1,2332 @@
...
+    in
+    SplitLiveRange.split v pv proc dom_functions rev_dom_functions
+    |> VariableRenaming.rename v bot_var dom_functions
+    |> DeadCodeElim.clean v bot_var
+
+  let ssify_proc ?splitting_strategy (proc : (Var.t, Program.e) Procedure.t)
+      (og_vars : Var.t Var.Decls.t) =
+    (* let all_vars = Procedure.local_decls proc in *)
+    let cfg =
+      match Procedure.graph proc with Some g -> g | None -> Procedure.G.empty
```
agle: If this returns Procedure.G.empty the next line will crash, it should just return proc as is if it has no graph and perform ssi only on procedures with a cfg defined.

PR #223, opened 2026-07-27
Smt backend

mira-alford:
Part 1 of the smt backend, yay!

Adds a summary inlining transform which replaces all procedure summaries with asserts/assumes.

Uses this in the dump-smt command, which outputs an smt file that can currently verify basic IL programs (pure and only using primitive types). Support for fancy things like datatypes/matching are not done yet. Should be structured to check that all assertions/ensures are satisfied.

Also makes some changes to the existing smt infrastructure. Fixes some bugs with incorrect symbols for bitvector operations, and makes a mess getting ITEs working (there has to be a better way??).


lib/lang/expr_smt.ml
```diff
@@ -378,12 +383,15 @@ module SMTLib2 = struct
     let* body = in_body in
     return @@ list [ quant; list binds; body ]
 
-  let smt_alg (e : sexp t BasilExpr.abstract_expr) =
+  let smt_alg' (e : sexp t BasilExpr.abstract_expr)
```
mira-alford: naming is hard, is this fine?

mira-alford:
Oops, should have predicted that failure 💀
Will fix.

lib/lang/expr_smt.ml
```diff
@@ -437,6 +445,24 @@ module SMTLib2 = struct
...
+              {
+                op = `Cases;
+                args = [ BinaryExpr { op = `IfThen; arg1; arg2 }; _ ];
+              },
+            [ _; arg3 ] ) ->
+            let* arg1 = arg1 in
+            let* arg2 = arg2 in
+            let* arg3 = arg3 in
+            return @@ list [ atom "ite"; arg1; arg2; arg3 ]
+        (* TODO actual matches. *)
```
agle: This would be simplified by using a pre-pass to rewrite cases expressions to an if-then-else intrinsic or function application expression. Rewrites in contrast to folds don't have difficulty with expanding and contracting multiple levels of the expression as they are always fixed -> unfixed n times -> fixed. We already do something similar to drop associative operators; a rewrite to compile cases to ite trees would be beneficial https://github.com/agle/bincaml/pull/223/changes#diff-aa74e48cc805b9e7b59ea1dde5eb351023184a3a0a26be9bc1f80c84ec069130R484. 

https://github.com/agle/bincaml/blob/eb2cd644d56d6346642661a61fee45c61110b437/lib/lang/expr.ml#L524

mira-alford: This is a way better solution, thanks for the idea!

Would you be fine with me expanding the IfThen op so that it can be intrinsic and treated as chained, or prefer a new op entirely?

lib/backends/smt.ml
```diff
@@ -0,0 +1,179 @@
...
+open Bincaml_util.Common
+open Bincaml_util
+open Expr_smt
+open Expr
+
+(* SMT Backend.
+   This backend expects the CFA Reduction and Summary Inlining transforms.
+   Will output a .smt file containing all declarations and asserts necessary
+   for verification.
+*)
```
katrinafyi: Here and in other module-level and function comments, using `**` will turn them into doc comments.

lib/lang/algsimp.ml
```diff
@@ -312,6 +312,36 @@ let algebraic_simplifications
...
+  match e with
+  | ApplyIntrin
+      {
+        op = `Cases;
+        args =
+          [
+            ( ( ApplyIntrin { op = `IfThen; args = [ cond; br_true ] },
+                Types.Boolean )
+            | ( BinaryExpr { op = `IfThen; arg1 = cond; arg2 = br_true },
+                Types.Boolean ) );
```
katrinafyi: It's my understanding that when `IfThen` is used as a case guard, it only appears inside BinaryExpr. Is this right?

If it's not needed, the match on <code>ApplyIntrin { op = `IfThen; ... }</code> can be deleted.

lib/lang/algsimp.ml
```diff
@@ -312,6 +312,36 @@ let algebraic_simplifications
       replace [%here] arg
   | _ -> Keep
 
+let if_then_else
+    (e :
+      (BasilExpr.t BasilExpr.abstract_expr * Types.t) BasilExpr.abstract_expr) =
```
katrinafyi, line 318: Just a doc comment about what this does and how to use it :)

lib/lang/ops.ml
```diff
@@ -298,7 +298,8 @@ module Spec = struct
           and catch fire *) ]
   [@@deriving show { with_path = false }, eq, ord]
 
-  type intrin = [ `Cases  (** choose first argument that is defined *) ]
+  type intrin =
+    [ `Cases | `IfThen  (** choose first argument that is defined *) ]
```
katrinafyi: If it's not too much trouble, I'd prefer to make a separate `IfThenElse` intrin rather than overloading the `IfThen` to two different meanings based on its number of arguments. 

Also note that the old doc comment was on the `Cases` variant, but after adding `IfThen` it becomes attached to the `IfThen` variant which isn't right.

lib/transforms/cfa_reduction.ml
```diff
@@ -82,14 +84,40 @@ let construct_final_edge proc =
...
+          | Stmt.Instr_Assert { body; attrib } ->
+              `Left
+                (Stmt.Instr_Assert
+                   {
+                     body =
+                       BasilExpr.binexp ~op:`IMPLIES
+                         (BasilExpr.rvar termination_var)
+                         body;
+                     attrib;
+                   })
```
katrinafyi: I generally like to avoid "pyramidal" structures of code. Especially in match expressions, it can make it hard to see where the pattern ends and the expression begins.

This is one suggestion to make it more compact by using a local variable.

```suggestion
          | Stmt.Instr_Assert { body; attrib } ->
              let body =
                BasilExpr.(binexp ~op:`IMPLIES (rvar termination_var) body)
              in
              `Left (Stmt.Instr_Assert { body; attrib })
```

lib/transforms/summary_inlining.ml
```diff
@@ -0,0 +1,89 @@
...
+    else block
+  in
+
+  (* Add ensures to return block. *)
+  if match List.head_opt return_id with Some bid -> true | _ -> false then
+    Block.append_stmts block
+      (List.map
+         (fun e -> Stmt.Instr_Assert { attrib = StringMap.empty; body = e })
+         spec.ensures)
+  else block
```
katrinafyi: `if match` sounds kind of Rusty (`if let`). Unfortunately, it's a bit unconventional in Ocaml. A simple match statement could be easier to read. You have also added `Procedure.get_return_block` so could you use that instead?

I'm also suspicious of the logic, it doesn't seem to do what the comments say it should do. It looks like if there exists an entry block, the code will add requires to `block`, but `block` may or may not be the entry block. Same for the ensures.

Do you need to check if `return_id == bid`? Or is there an underlying assumption here?

lib/invariants.ml
```diff
@@ -13,11 +13,13 @@ type t =
...
   | NoPhis
+  | NoSummaries
   | Params
   | LambdaLift
   | MemoryEncoding
   | GtirbArm
   | ReducibleLoops
+  | Acyclic
       (** All loops are reducible. That is, there are no {i irreducible} loops.
       *)
```
katrinafyi, line 23: the doc comment should stay on `ReducibleLoops`, rather than `Acyclic`.

bin/main.ml
```diff
@@ -167,6 +167,26 @@ let repl ~verb ~echo_cmd =
...
+    | Ok (`Atom "live-smt" :: fnames as l) ->
+        let c = last fnames in
+        let l =
+          List.take (List.length l - opt_len c) l
+          |> List.to_string ~sep:" " CCSexp.to_string
+        in
+        (match c with Some n -> Iter.singleton n | None -> Iter.empty)
+        |> Iter.flat_map complete_filename
+        |> Iter.map (fun s -> l ^ " " ^ s)
+        |> Iter.iter (LNoise.add_completion completions)
```
katrinafyi: As best as I can tell, these two match arms are identical and could be combined using `|`.

katrinafyi (commented):
Pretty cool!

lib/transforms/summary_inlining.ml
mira-alford: Very true on that second point! I accidentally put bid in the pattern thinking it would match where they are equal instead of shadowing lol. It was causing some problems as you can imagine.

lib/lang/ops.ml
mira-alford: Good idea

lib/backends/smt.ml
```diff
@@ -0,0 +1,257 @@
...
+       ~eq:(fun v1 v2 -> String.equal (Var.name v1) (Var.name v2))
+  |> Iter.map VarSet.of_list
+  (* Only keep lists of length > 1: *)
+  |> Iter.filter (VarSet.cardinal %> ( <= ) 2)
+
+(* Map rvars to sexps, necessary to handle ambiguities and
+   function calls which are sensitive to context in program. *)
+let rvar_map (program : Program.t) =
+  ambiguities program
+  (* Map all ambiguous variables to as expressions. *)
```
agle, line 41: I may be misunderstanding but is this to deal with locals that share names with different types in different procedures? If so could you explain briefly in a comment please.

Is it dealing with local variables across different procedures? IMO it would be safer to always mangle local variables with the procedure name, coincident names that are not defined locally in different procedures would constitute an malformed program (I believe there is a check for this but I may be wrong). Sharing definitions of local variables with the same name across different procedures seems to risk funny semantics- creating invalid constraints between locals of different procedures.

For verifying a single procedure at a time, I think you could define locals after the `(push)` and they will be discarded after the `(pop)`; but without mangling this could still restrict how easy it is to share subgoals between different procedure checks (it _requires_ you pop before analysing another procedure).

Kait suggested it might be to do with inlining definitinos of types where variables are only annotated with their name; in that case it doesn't seem safe to include all local variables in the one map; they might just have different types in different procedrues?

lib/backends/smt.ml
```diff
@@ -0,0 +1,257 @@
...
+
+  CCVector.push builders (builder, ctx);
+
+  (* Translate each statement to smt. *)
+  Procedure.iter_stmt_topo_fwd procedure
+  |> flip Iter.for_each (fun stmt ->
+      let builder = SMTLib2.empty in
+      let ctx = { ctx with stmt = Some stmt; vc = false } in
+      match stmt with
+      | Stmt.Instr_Assert { body } ->
```
agle: This is a procedure post-CFA reduction right? So you basically have one block containing a list of assumes/assigns and ending in an assert? So assert is basically the final reachability condition? It would be neater to represent the verification condition in a custom datastructure; basically a hoare triple implication P { Program } Q. Where P is precond, { Program} is the passified set of assumes and assignments and Q is the safe reachability condition for `return`. Aaron's CHC backend is similar to this, forced by the structure of CHCs.

lib/transforms/cfa_reduction.ml
```diff
@@ -82,34 +84,57 @@ let construct_final_edge proc =
...
+    let proc, id = Procedure.fresh_block proc ~stmts:[] () in
+    let proc =
+      Procedure.modify_block proc id (fun b -> { b with stmts = final_edge })
+    in
+
+    (* Make this the entry and return block. *)
+    let proc = Procedure.set_entry_block proc id in
+    Procedure.PG.map_graph
+      (fun g ->
+        Procedure.G.add_edge g (Procedure.Vert.End id) Procedure.Vert.Return)
```
agle, line 139: I think an intermediate type would be simpler, rather than proc with flattened graph e.g. `{reduced_proc: Stmt.t list, reach_end_pred: Expr.t list;}`), the caller can combine with the procedure `id` or copy of the procedure if needed.

lib/transforms/cfa_reduction.ml
```diff
@@ -82,34 +84,57 @@ let construct_final_edge proc =
...
+      let assert_stmts, non_guard_stmts =
+        non_guard_stmts
+        |> List.partition_filter_map (function
+          | Stmt.Instr_Assert { body; attrib } ->
+              let body =
+                BasilExpr.(binexp ~op:`IMPLIES (rvar termination_var) body)
+              in
+              `Left (Stmt.Instr_Assert { body; attrib })
+          | other -> `Right other)
+      in
```
agle, line 99: lgtm

lib/backends/smt.ml
```diff
@@ -0,0 +1,257 @@
...
+              let typ = fst @@ SMTLib2.of_typ (Var.typ v) in
+              (CCSexp.(list [ atom "as"; var; typ ]), s)
+            in
+            (v, sexp))
+  |> VarMap.of_iter
+
+(* Produce a list of builders for a procedure, with context.
+   Builder for local declarations + one for each statement.
+   Assertions produce a second builder for verification. *)
+let build_procedure ~rvars (program : Program.t) (procedure : Program.proc) :
```
agle, line 57: This could maybe be more efficient by passing a iter callback function (`(builder * ctx) -> unit`) which you call instead of `CCVector.push`. This means you could construct an `Iter.t` from this function to pass each intermediate check to the solver.

lib/lang/expr_smt.ml
```diff
@@ -357,6 +381,12 @@ module SMTLib2 = struct
     | `MapAccess -> atom "select"
     | `MapUpdate -> atom "store"
     | `IMPLIES -> atom "=>"
+    | `INTADD -> atom "+"
+    | `INTMUL -> atom "*"
+    | `INTSUB -> atom "-"
+    | `INTDIV -> atom "/"
```
ncough, line 387: Probably should be `div`, `/` is for reals.

lib/transforms/cfa_reduction.ml
```diff
@@ -82,34 +84,57 @@ let construct_final_edge proc =
...
+          4. any assertions from the block *)
       CCVector.push final_edge ites;
       CCVector.append_list final_edge non_guard_stmts;
       CCVector.push final_edge termination;
+
+      (* Asserts are moved to the end.
+         They should be order invariant thanks to SSA,
+         however the added dependence on the termination
+         condition requires them after it's declared. *)
+      CCVector.append_list final_edge assert_stmts;
```
ncough, line 114: I am just glancing through, so maybe completely wrong. Is it safe here to reorder asserts and assumes? Or are assumes fixed to entry?

lib/transforms/type_check.ml
```diff
@@ -196,6 +196,17 @@ let type_check stmt_id block_id expr =
...
+    | `IfThenElse -> (
+        match args with
+        | [ Types.Boolean; arg1; arg2 ] ->
+            if Types.equal arg1 arg2 then []
+            else
+              [
+                type_err "non-equal ite branches : %s %s" (Types.to_string arg1)
+                  (Types.to_string arg2);
+              ]
+        | [ Types.Boolean; arg1 ] -> []
```
ncough, line 208: I'm probably missing something, but this case doesn't make a lot of sense to me. What does the single argument case return if the condition is false?

lib/lang/ops.ml
```diff
@@ -298,7 +298,8 @@ module Spec = struct
           and catch fire *) ]
   [@@deriving show { with_path = false }, eq, ord]
 
-  type intrin = [ `Cases  (** choose first argument that is defined *) ]
+  type intrin =
+    [ `Cases | `IfThenElse  (** choose first argument that is defined *) ]
```
ncough, line 302: Is the comment on the wrong variant?

lib/transforms/remove_loops.ml
```diff
@@ -0,0 +1,117 @@
...
+  let header_block = Procedure.get_block proc header |> Option.get in
+
+  let dest = ProcIntra.BlockGraph.E.dst in
+  let src = ProcIntra.BlockGraph.E.src in
+
+  (* Treat any assertions in the header as invariants. *)
+  let invariants =
+    Block.fold_forwards ~phi:const
+      ~f:(fun acc stmt ->
+        match stmt with Stmt.Instr_Assert { body } -> body :: acc | _ -> acc)
```
ncough, line 31: Will this include asserts after a statement? Not sure that's right.

lib/transforms/remove_loops.ml
```diff
@@ -0,0 +1,117 @@
...
+      [] header_block
+  in
+
+  (* Add havoc statements. *)
+  let havocs =
+    Block.free_vars header_block
```
ncough, line 37: Is this the variables mentioned in the block, or the live variables across it? Should be the latter. I guess SSA should guarantee these are the same, but does this pass assume SSA?

lib/transforms/remove_loops.ml
```diff
@@ -0,0 +1,117 @@
...
+
+  (* Assume the invariant after havocing. *)
+  let assumes =
+    invariants
+    |> List.map (fun inv ->
+        Stmt.Instr_Assume { body = inv; attrib = Attrib.empty; branch = false })
+  in
+
+  (* Update header with new statements. *)
+  let header_block = Block.append_stmts header_block (havocs @ assumes) in
```
ncough, line 57: This places havoc and assume after loop header statements? I'd assume they go just after the asserts.

ncough (commented):
I'm not quite done reading through yet, but in the interest of not holding anything up, here are some things I have found. I think most are me just not knowing bincaml :)

agle (commented):
Nick's already looked at the loop stuff so I'll just comment on this part

lib/transforms/remove_loops.ml
```diff
@@ -0,0 +1,117 @@
...
+  let proc = Procedure.update_block proc header header_block in
+
+  (* Add invariants to a backedge, subbing phi node vars. *)
+  let add_invariants =
+    let phis =
```
ncough, line 62: Again, probably confused, but I would have thought you'd want to substitute phis based on what this edge is going to do. Here it seems to be substituting for header self loops?

lib/transforms/remove_loops.ml
```diff
@@ -0,0 +1,117 @@
...
+  let loops =
+    block_infos
+    |> List.filter (fun block ->
+        match ProcIntra.classify_block block with
+        | `ReducibleHeader -> true
+        | _ -> false)
+  in
+  List.fold_left (transform_loop prog) proc loops
+
+let transform (prog : Program.t) =
```
ncough, line 116: Just wanted to comment that I find this implementation quite clean and elegant. Doing something rather intricate in ~100 lines.

lib/transforms/remove_loops.ml
```diff
@@ -0,0 +1,117 @@
...
+  The only difference is that we put asserts at top of loop header block
+  instead of at the ends of non-entry blocks. *)
+
+let transform_loop (prog : Program.t) (proc : Program.proc)
+    (loop : ProcIntra.block_info) =
+  let header, next_h, headers, nodes, entries, backedges =
+    match loop with
+    | { block; loop = PrimaryHeader { primary_header; headers; nodes } } ->
+        let entries = ProcIntra.compute_entries proc loop in
+        let backedges = ProcIntra.compute_backedges proc loop in
```
ncough, line 18: Is there any issue with calling these after altering `proc`? Maybe given a nested loop or something?

lib/transforms/summary_inlining.ml
```diff
@@ -0,0 +1,91 @@
...
+          (List.map
+             (fun e ->
+               Stmt.Instr_Assume
+                 { attrib = StringMap.empty; body = e; branch = false })
+             spec.requires)
+    | _ -> block
+  in
+
+  (* Add ensures to return block. *)
+  match List.head_opt return_id with
```
ncough, line 75: Is there an invariant ensuring at most one of these?

lib/transforms/summary_inlining.ml
```diff
@@ -0,0 +1,91 @@
...
+                  Stmt.Instr_Assert
+                    { attrib = StringMap.empty; body = subst_expr args e })
+            in
+            let ensures =
+              spec.ensures
+              |> List.map (fun e ->
+                  Stmt.Instr_Assume
+                    {
+                      attrib = StringMap.empty;
+                      body = subst_var lhs @@ subst_expr args e;
```
ncough, line 48: A staged subst like this has caught me before. Can you have something like: `callee(x): y ensures y = x + 1`, invoked as `z := callee(y)`? Ends up with `assume (z = z + 1)`.

lib/backends/smt.ml
mira-alford: It is dealing with variables across procedures as all declarations need to happen globally (cannot be inside a push/pop scope). It is quite an annoying limitation but the smtlib spec suggests (when generating smt code) to just disambiguate types using as expressions and so long as all asserts are inside scopes (this is the case) procedures should never interfere. I am a fan of the renaming idea if you think that would be better though, I'm not the biggest fan of this approach.

lib/backends/smt.ml
mira-alford: Is this comment about performance or nicer code? I think I agree from our conversation the other day that it is better regardless and am implementing it but i just remembered that it still requires the entire list before it can verify individual blocks (this could be worked around with some effort, I'm contemplating it 😄 ).

lib/transforms/remove_loops.ml
mira-alford: Thanks!

lib/transforms/remove_loops.ml
mira-alford: It does assume SSA... except that is not listed in the passes 👀

lib/backends/smt.ml
agle: If the constraints themselves are scoped then there's no problem. Its not obvious whether mangling variables is going to be more or less convoluted, so up to you if you want to implement it that way.

lib/backends/smt.ml
mira-alford: Very big changes to how the smt backend file is structured, but they have cleaned it up a *lot* and I am pleased with the result. It is a little broken for now though, need to be able to easily order declarations by dependency. lmk what you think.

lib/transforms/type_check.ml
mira-alford: Very true, i think this was copied erroneously from the old if then.

lib/transforms/remove_loops.ml
mira-alford: It shouldn't have any other statements, except optionally SSA variable declarations which we would want before the asserts anyway.

lib/transforms/cfa_reduction.ml
mira-alford: It assumes they are all guards at the entry, added a comment explaining this. Dealing with mixed in assumes was very difficult, trying to remember why but I'll think about it.

lib/transforms/remove_loops.ml
mira-alford: Oops, fixed.

lib/transforms/remove_loops.ml
mira-alford: Each loop header should be unique and it only modifies the header + edges back to that header, so I'm fairly sure it should be fine? none of that will modify other IDs.

lib/transforms/summary_inlining.ml
mira-alford: Fixed.

lib/transforms/summary_inlining.ml
mira-alford: Ah wonderful...

mira-alford: Fixed!

lib/transforms/cfa_reduction.ml
mira-alford: The only thing returned is a stmt list so it is unclear why we would need to wrap that in a struct? Can just call construct_final_edge if you want the statement list directly (admittedly as a vector).

lib/transforms/cfa_reduction.ml
mira-alford: This is a problem, for now we just assume there are only guards. I don't think it should be solved in this PR, but instead would be solved by adding an SSI transform.

PR #225, opened 2026-07-27
ci: build nix packages and dev shell in ci

### LINKED ISSUE #209: Run Nix in CI
I've been sloppy in maintaining the Nix through recent package and dependency changes.

Running Nix in CI would check that we don't break the Nix users. We would probably want to make it an optional CI check.

katrinafyi:
Closes https://github.com/agle/bincaml/issues/209

As long as we don't mark it as a Required check, it will still be possible to merge PRs without the Nix build succeeding. This just makes it much more visible if it does break.


mira-alford (approved):
Nice!

PR #226, opened 2026-07-27
nix: Use Dune 3.24 in bincaml build

katrinafyi:
Surprisingly, that wasn't too bad...

This lets us use new dune features without having to wait for nixpkgs to roll the version. The new Dune is used in bincaml only and the rest of the ocaml packages still use the nixpkgs default to avoid rebuilding everything.


mira-alford:
Nice!

Does it bump the shells dune up too?

katrinafyi:
Good question, it probably has both versions in there, idk which takes precedence..

katrinafyi:
Seems to work :D 
```bash
$ nix develop .#no-fp -L
warning: Git tree '/home/rina/progs/obasil' is dirty
[rina@leng obasil]$ which dune
/nix/store/8s0x8xb0ykrcensoredp5as8kvpxrilx-dune-3.24.1/bin/dune
```

mira-alford (approved):
Another day, slightly more diverging from nixpkgs. This all works for me, looks good.

PR #227, opened 2026-07-30
hm: avoid `include` and call `fix` in smart constructors 

katrinafyi:
This removes the uses of include which accumulated all earlier HM modules into later ones. This makes it so each function has one canonical definition site, and avoids inflating the docs.

For simplicity, this is just done by changing includes to opens, so it is still a little hard to see where a function comes from. But I think it is still an improvement.

Independently, this also embeds calls to `fix` into the type smart constructors. This avoids needing to call fix separately after every call to the constructors, and is planned to help with moving the HC state to value level. This might have a side-effect of eagerly adding all the primitive types to every HC context, but I think this is not a big problem.

These two changes could be separated if you want.

follow up to https://github.com/agle/bincaml/pull/216

### DIFF
--- a/lib/lang/hm/unification.ml
+++ b/lib/lang/hm/unification.ml
@@ -3,11 +3,10 @@
 open Common
 open Abstract_expr
 
-module Make
-    (T : TypeExpr.TypeContext)
-    (Hm_types' : module type of Hm_types.Make (T)) =
-struct
-  open Hm_types'
+module Make (T : TypeExpr.TypeContext) = struct
+  open Hm_types.Make (T)
+  (* XXX: todo *)
+
   open T
   open T.Typ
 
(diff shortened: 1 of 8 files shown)


lib/lang/hm/unification.ml
```diff
@@ -4,28 +4,29 @@ open Common
 open Abstract_expr
 
 module Make (T : TypeExpr.TypeContext) = struct
-  open TypeExpr
-  include T
-  include Typ
-  open Hm_types
-  include Hm_types.Make (T)
+  open Hm_types.Make (T)
+  (* XXX: todo *)
```
agle, line 8: boop

agle (approved):
This is good. Idk why it works I thought I had some trouble with functor type escaping the functor and needing to store the parameter in the result, and therefore becoming a distinct type to the other modules e.g. A.T.t != B.T.t?

Calling fix early shouldn't be a problem

katrinafyi:
Yeah I did see that error when working on it. I think it's when you return a type which is defined in the functor argument, and it was happening because of the includes. 

In this PR, I think it's able to unify all of the types after instantiating the functors?

lib/lang/hm/unification.ml
katrinafyi: oops I forgot this sorry

PR #229, opened 2026-07-30
aslp: merge `$mem` functionality into `add_aarch64_global_declarations` 

katrinafyi:
This makes it use the same global variable declaration logic as the normal registers. This simplifies the code a bit and it avoids inserting memory if none of the lifted instructions reference it.

After this change, we can remove `bincaml_memory_var` from the IBI parameters entirely.

### DIFF
--- a/lib/transforms/aslp/aslp.ml
+++ b/lib/transforms/aslp/aslp.ml
@@ -142,21 +142,24 @@ let transform_procedure proc =
 
 (** Adds architectural variable declarations to the given program. By default,
     includes only those variables which are used. *)
-let add_aarch64_global_declarations ?(add_all = false) prog =
-  let include_var =
-    if add_all then Fun.const true
+let add_aarch64_global_declarations ?(include_unused = false) prog =
+  let include_predicate =
+    if include_unused then Fun.const true
     else
       Fun.flip VarSet.mem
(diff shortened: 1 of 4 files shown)


lib/transforms/aslp/aslp.ml
```diff
@@ -187,12 +162,23 @@ let add_aarch64_global_declarations ?(add_all = false) prog =
...
+  (match Program.get_decl_by_name (Aslp_lexpr.name Memory) prog with
+  | Some (Variable { binding }) ->
+      if not (Types.equal (Var.typ binding) (Aslp_lexpr.typ Memory)) then
+        Logs.warn (fun m ->
+            m
+              "Memory declared with unexpected type; lifter may produce \
+               wrongly-typed or incorrect code.")
+  | Some _ ->
+      failwith (Aslp_lexpr.name Memory ^ " already declared as non-variable")
+  | None -> () (* will be added by add_aarch64_global_declarations *));
```
agle: This doesn't seem neccessary anymore? Or it would be wanted for all of the variables?

agle (approved):

lib/transforms/aslp/aslp.ml
katrinafyi: Fair cop.

PR #230, opened 2026-08-04
`nix develop`-based fast path in CI

katrinafyi:
i am back on the nix opam shell grind. i want it to give a fast path for ci which is less maintenance steps and less flaky than the docker image. it should still support falling back to opam install in case the nix shell is missing some packages, so that a failing nix build doesn't block CI.

i'm abandoning my approach from last week, because that approach was using opam install --fake which is "dangerous and likely to break your opam environment". the goal of that was to somehow "teach" opam about the nixpkgs ocaml packages so opam will only install the missing ones, but i think this would be very fragile and didn't get to a proof of concept stage.

instead, i think it can be something much simpler. you just use all nixpkgs ocaml packages and if that fails, then use all opam packages. it's a lot slower in the fallback case, but hopefully that should be rare enough that it doesn't matter.

future work (beyond today), will look at caching the nixpkgs packages in github actions

- [ ] port other jobs too 
- [ ] gate thingy behind all nix jobs succeeding

### DIFF
--- a/.github/workflows/test-using-nix.yml
+++ b/.github/workflows/test-using-nix.yml
@@ -66,3 +66,6 @@ jobs:
     if: >
       needs.build_test.result == 'success' || needs.build_test_slower.result == 'success'
 
+    steps:
+      - run: true
+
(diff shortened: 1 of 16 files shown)


katrinafyi:
Abandoning. Related:
- https://github.com/UQ-PAC/bincaml/pull/230
- https://github.com/UQ-PAC/bincaml/pull/239
- https://github.com/UQ-PAC/bincaml/pull/243

PR #231, opened 2026-08-07
downgrade `lang dune` back to 3.19 until we use the new features

katrinafyi:
although the nix uses Dune 3.24 now, we don't use any of the new features yet so it's more convenient to keep the lower dune requirement for now. this was causing me minor inconvenience on my home laptop which has older dune.


agle (approved):

katrinafyi:
Re https://github.com/agle/bincaml/pull/226

PR #232, opened 2026-08-07
nix: add `.#bincamlDocs.docs` to run `odoc_driver` using nix

katrinafyi:
this turns the documentation site into a Nix package, so we can (eventually) simplify the CI/CD for doc generation.

you can test it with:

    nix build .#bincamlDocs.docs -Lv

at the moment, it just builds the docs for the `fmt` package to keep it fast and small, but when it's done, it will build all the bincaml docs - same as the current github workflow.

this involved packaging some odoc packages which were missing from nixpkgs. it also involves some careful folder structure construction and some faking, because the `odoc_driver` is built around the assumption of having opam switches.

at the moment, it all works *aside from* it miscategorises the builtin packages (like `unix`) as separate opam packages.

future work: sherlodoc

this PR also adjusts the `./.` sources to all reference the bincaml source, and changes the shell.nix so that more dependencies are properly specified in the package definitions so fewer packages need to be explicitly mentioned in the shell.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,4 +1,9 @@
 {
+  nixConfig.extra-substituters = [ "https://pac-nix.cachix.org/" ];
+  nixConfig.extra-trusted-public-keys = [
+    "pac-nix.cachix.org-1:l29Pc2zYR5yZyfSzk1v17uEZkhEw0gI4cXuOIsxIGpc="
+  ];
+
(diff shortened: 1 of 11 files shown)


mira-alford (approved):
Very nice cleanups, was a headache adding all the packages everywhere. This all works for me.

PR #236, opened 2026-08-14
Store program declaration order explicitly

### LINKED ISSUE #235: Declaration order
Declaration order is currently strictly in the order of ID creation, this works for parsing a file in order but is bad for constructing custom IR directly where IDs are created out of order.

Constraints:

- O(n) iteration
- O(1) lookup by ID.t (?) (better than O(n) at least)  `program -> ID -> declaration` lookup
    - probably not crucial if we always modify by a map over the decl list if we are modifying a large chunk of the declarations?
- immutable/persistent data structure

Proposal

Set of categorised lists

```ocaml
type decls = {
   type_decls : declaration list;
   variable_decls: declaration list;   
   function_decls: declaration list;
   procedure_decls: declaration list;
}

let declarations {type_decls; variable_decls; function_decls; procedure_decls} = type_decls @@ variable_decls @@ function_decls @@ procedure_decls
let get_decl prog id = memoised (declarations |> List.map (fun d -> decl_id d, d) |> IDMap.of_iter)
```

Just a list
- hard to do random insertion e.g. 'before all procs'
- list concatenation inefficient

```ocaml
type decls = declaration list;
let insert_before_procs prog decl = let index = List.index_of (function Procedure _ -> true | _ -> false) in List.take index prog.decls @@ decl @@ List.drop index prog.decls
```

 Or a random access persistent vector type
- We don't have a good library right now

- Or a CCVector 
    - good for random access
    - not great for persistent ops

---

~~Of these a simple association list is probably the most reasonable for simplicity of specific order, and predictable (if non-ideal) performance behaviour. More complex solutions probably reduce to just memoising functions over an association list to speed up access, which we can do if needed.~~ assoc lists don't preserve order--would have to do something custom and incur concatenation cost for insert in middle :(

https://uq-pac.github.io/bincaml/containers/containers/CCList/index.html#association-lists_2

chunk sequences is probably better, with linear lookup: https://ocaml.org/p/sek/20260619/doc/sek/Sek/Persistent/index.html

- still linear lookup by id

---

- [ ] Boogie backend currently explicitly chooses a declaration order to emit boogie decls  to fix the dependencies---following this change we can update it to create a program with declarations in the right order instead.

agle:
Fixes #235 

Declaration order is now defined by the order of operations over `program` via Program.add_decl (an immutable structure), rather than the order in which ID.fresh is called (which is side-effecting).

 `Program.add_decl` now takes the optional parameter `?at`, which allows inserting at a specific position in the declaration order.

```ocaml
(** [add_decl ?at ~attrib prog decl] adds [decl] to the declaration list of
[prog]. [?at] defaults to [`Append] which is constant time. Note that all other
values of [at] take at least linear time. *)
let add_decl
    ?(at : [ `Append | `Prepend | `BeforeVars | `BeforeFuncs | `BeforeProcs ] =
      `Append) ?(attrib = StringMap.empty) p decl =
```

Implementation

- util/indexed_list.ml: defines a map structure with a list defining its iteration order `'a t = {order: KEY.t List.t; values: 'a Map.Make(KEY)}`. The performance characteristics could relatively easily be improved by changing the linear datastructure. 
- moves exported types in `Program` shared by mli and ml files to a separate ml file.
- moves `implicit_declaration` (which defines variant constructors) into a variant of `declaration` so that everything is in one list.
- aslp frontend now puts global variable declarations at the start of the program.
- some minor resulting simplifications of `loadir.ml`

### DIFF
--- a/lib/lang/program.ml
+++ b/lib/lang/program.ml
@@ -27,10 +27,6 @@ let decl_is_typ = function Type _ -> true | _ -> false
 let decl_is_func = function Function _ -> true | _ -> false
 let decl_is_var = function Variable _ -> true | _ -> false
 let decl_is_proc = function Procedure _ -> true | _ -> false
-let types_before _ k = not (decl_is_typ k)
-let vars_before _ k = decl_is_proc k || decl_is_func k || decl_is_var k
-let procs_before _ k = decl_is_proc k
-let funcs_before _ k = decl_is_func k
 
 let decl_binding = function
   | Type { binding } -> binding
@@ -183,6 +179,9 @@ let get_id_by_name name prog = prog.global_names.get_id name
 let add_decl
     ?(at : [ `Append | `Prepend | `BeforeVars | `BeforeFuncs | `BeforeProcs ] =
       `Append) p decl =
+  let vars_before _ k = decl_is_proc k || decl_is_func k || decl_is_var k in
+  let procs_before _ k = decl_is_proc k in
+  let funcs_before _ k = decl_is_func k in
   let d = p.global_names.decl_or_get (decl_binding decl) in
   let declarations =
     match at with
--- a/lib/lang/program.ml
+++ b/lib/lang/program.ml
@@ -2,11 +2,7 @@ open Common
 open Types
 open Expr
 open Containers
-
-type e = BasilExpr.t
-type proc = (Var.t, BasilExpr.t) Procedure.t
-type bloc = (Var.t, BasilExpr.t) Block.t
-type stmt = (Var.t, Var.t, e) Stmt.t
+include Program_types
 
 module Proc = struct
   type t = proc
@@ -25,35 +21,23 @@ let show_stmt =
 
 let pp_stmt fmt s = Format.pp_print_string fmt (show_stmt s)
 
-type prog_spec = { rely : BasilExpr.t list; guarantee : BasilExpr.t list }
-type func_type = Axiom of e | Uninterpreted | Function of e
+module DeclsList = Bincaml_util.Indexed_list.Make (ID)
 
-type implicit_declaration =
-  | VariantCase of {
-      variant : string;
-      belongs_to : Types.t;
-      constructor : Var.t; (* function to construct a value of this case *)
-    }
-
-type declaration =
-  | Type of { binding : string; typ : Types.t }
-  | Function of {
-      binding : Var.t;
-      attrib : Attrib.attrib_map;
-      definition : func_type;
-    }  (** pure functions *)
-  | Variable of {
-      binding : Var.t;
-      attrib : Attrib.attrib_map;
-      classification : e option;
-    }
-  | Procedure of { definition : proc }
+let decl_is_typ = function Type _ -> true | _ -> false
+let decl_is_func = function Function _ -> true | _ -> false
+let decl_is_var = function Variable _ -> true | _ -> false
+let decl_is_proc = function Procedure _ -> true | _ -> false
+let types_before _ k = not (decl_is_typ k)
+let vars_before _ k = decl_is_proc k || decl_is_func k || decl_is_var k
+let procs_before _ k = decl_is_proc k
+let funcs_before _ k = decl_is_func k
 
 let decl_binding = function
   | Type { binding } -> binding
   | Variable { binding } -> Var.name binding
   | Function { binding } -> Var.name binding
   | Procedure { definition } -> ID.to_string (Procedure.id definition)
+  | Implicit (VariantCase _) -> failwith "variant case has no binding :("
 
 let pretty_proc p =
   let show_lvar v = Containers_pp.text @@ Var.to_string_il_lvar v in
@@ -64,6 +48,7 @@ let pretty_proc p =
 let pretty_declaration d =
   let open Containers_pp in
   match d with
+  | Implicit _ -> failwith "unsupp"
   | Variable { binding; attrib; classification } ->
       let classification =
         classification |> Option.to_list
@@ -102,8 +87,7 @@ let pretty_declaration d =
 
 type t = {
   modulename : string;
-  declarations : declaration IDMap.t;
-  implicit_decls : implicit_declaration IDMap.t;
+  declarations : declaration DeclsList.t;
   entry_proc : ID.t option;
   global_names : ID.generator;
   attrib : Attrib.attrib_map;
@@ -121,7 +105,7 @@ let entry_proc_exn p =
   p.entry_proc |> function
   | None -> raise Not_found
   | Some i -> (
-      IDMap.get i p.declarations |> function
+      DeclsList.get i p.declarations |> function
       | Some (Procedure { definition }) -> definition
       | _ -> raise Not_found)
 
@@ -132,32 +116,32 @@ let map_procedures f p =
     p with
     declarations =
       p.declarations
-      |> IDMap.mapi (fun i -> function
+      |> DeclsList.mapi (fun _ i -> function
         | Procedure { definition } -> Procedure { definition = f i definition }
         | o -> o);
   }
 
 let proc prog p =
-  IDMap.find p prog.declarations |> function
+  DeclsList.find p prog.declarations |> function
   | Procedure { definition } -> definition
   | _ -> raise Not_found
 
 let proc_opt prog p = try Some (proc prog p) with Not_found -> None
 
 let procs p =
-  IDMap.to_iter p.declarations
+  DeclsList.to_iter p.declarations
   |> Iter.filter_map (function
     | k, Procedure { definition } -> Some (k, definition)
     | _ -> None)
 
 let global_vars prog =
-  IDMap.values prog.declarations
+  DeclsList.values prog.declarations
   |> Iter.filter_map (function
     | Variable { binding } -> Some binding
     | _ -> None)
 
 let global_constants prog =
-  IDMap.values prog.declarations
+  DeclsList.values prog.declarations
   |> Iter.filter_map (function
     | Function { binding } -> Some binding
     | _ -> None)
@@ -165,10 +149,10 @@ let global_constants prog =
 let get_decl_by_name_id name prog =
   try
     let id = prog.global_names.get_id name in
-    IDMap.find_opt id prog.declarations |> Option.map (fun v -> (id, v))
+    DeclsList.find_opt id prog.declarations |> Option.map (fun v -> (id, v))
   with Not_found -> None
 
-let get_decl id prog = IDMap.find_opt id prog.declarations
+let get_decl id prog = DeclsList.find_opt id prog.declarations
 let get_decl_by_name name prog = get_decl_by_name_id name prog |> Option.map snd
 
 let get_proc id prog =
@@ -184,20 +168,38 @@ let get_proc_by_name name prog =
 let get_implicit_decl_by_name name prog =
   try
     let id = prog.global_names.get_id name in
-    IDMap.find_opt id prog.implicit_decls
+    DeclsList.find_opt id prog.declarations |> function
+    | Some (Implicit i) -> Some i
+    | _ -> None
   with Not_found -> None
 
 let declare_name_exn name prog = prog.global_names.decl_exn name
 let declare_name name prog = prog.global_names.decl_or_get name
 let get_id_by_name name prog = prog.global_names.get_id name
 
-let add_decl ?(attrib = StringMap.empty) p decl =
+(** [add_decl ?at ~attrib prog decl] adds [decl] to the declaration list of
+    [prog]. [?at] defaults to [`Append] which is constant time. Note that all
+    other values of [at] take at least linear time. *)
+let add_decl
+    ?(at : [ `Append | `Prepend | `BeforeVars | `BeforeFuncs | `BeforeProcs ] =
+      `Append) ?(attrib = StringMap.empty) p decl =
   let d = p.global_names.decl_or_get (decl_binding decl) in
-  { p with declarations = IDMap.add d decl p.declarations }
+  let declarations =
+    match at with
+    | `Append -> DeclsList.append d decl p.declarations
+    | `Prepend -> DeclsList.prepend d decl p.declarations
+    | `BeforeFuncs ->
+        DeclsList.insert_before ~before:funcs_before d decl p.declarations
+    | `BeforeVars ->
+        DeclsList.insert_before ~before:vars_before d decl p.declarations
+    | `BeforeProcs ->
+        DeclsList.insert_before ~before:procs_before d decl p.declarations
+  in
+  { p with declarations }
 
 let remove_decl p decl =
   let d = p.global_names.decl_or_get (decl_binding decl) in
-  { p with declarations = IDMap.remove d p.declarations }
+  { p with declarations = DeclsList.remove d p.declarations }
 
 let update_decl ?(attrib = StringMap.empty) prog decl =
   add_decl ~attrib prog decl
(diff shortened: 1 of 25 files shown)


lib/lang/program.ml
```diff
@@ -64,6 +48,7 @@ let pretty_proc p =
 let pretty_declaration d =
   let open Containers_pp in
   match d with
+  | Implicit _ -> failwith "unsupp"
```
katrinafyi, line 47: this must be printable for debugging reasons, even if it cannot be parsed back.

lib/lang/program.ml
```diff
@@ -184,20 +168,38 @@ let get_proc_by_name name prog =
...
 let declare_name name prog = prog.global_names.decl_or_get name
 let get_id_by_name name prog = prog.global_names.get_id name
 
-let add_decl ?(attrib = StringMap.empty) p decl =
+(** [add_decl ?at ~attrib prog decl] adds [decl] to the declaration list of
+    [prog]. [?at] defaults to [`Append] which is constant time. Note that all
+    other values of [at] take at least linear time. *)
+let add_decl
+    ?(at : [ `Append | `Prepend | `BeforeVars | `BeforeFuncs | `BeforeProcs ] =
+      `Append) ?(attrib = StringMap.empty) p decl =
```
katrinafyi: attrib unused. add "or update".
```suggestion
(** [add_decl ?at ~attrib prog decl] adds [decl] to the declaration list of
    [prog], or updates it without re-ordering if it already exists.
    [?at] defaults to [`Append] which is constant time. Note that all
    other values of [at] take at least linear time. *)
let add_decl
    ?(at : [ `Append | `Prepend | `BeforeVars | `BeforeFuncs | `BeforeProcs ] =
      `Append) p decl =
```

lib/lang/program.ml
```diff
@@ -184,20 +168,38 @@ let get_proc_by_name name prog =
...
+  in
+  { p with declarations }
 
 let remove_decl p decl =
   let d = p.global_names.decl_or_get (decl_binding decl) in
-  { p with declarations = IDMap.remove d p.declarations }
+  { p with declarations = DeclsList.remove d p.declarations }
 
 let update_decl ?(attrib = StringMap.empty) prog decl =
   add_decl ~attrib prog decl
```
katrinafyi: I would personally not have two names for same thing. Someone might expect that `update_decl` would error if it doesn't previously exist, but it will actually automatically add it. 

`update_decl` can just be deleted imo

lib/lang/program_types.ml
```diff
@@ -0,0 +1,35 @@
...
+    }  (** pure functions *)
+  | Variable of {
+      binding : Var.t;
+      attrib : Attrib.attrib_map;
+      classification : e option;
+    }
+  | Procedure of { definition : proc }
+  | Implicit of implicit_declaration
+      (** A declaration which is not explicitly present in the ir program, but
+          created by another construct, e.g. type constructors. *)
```
katrinafyi, line 35: Can you help me to understand why the new case is needed? Could you not instead, for example, treat the `Type` case as declaring all its constructors as well as the type name? Or would this be less preferable?

lib/transforms/gamma_vars.ml
```diff
@@ -35,7 +35,7 @@ let add_globals ?(check_names = false) (add : Var.t -> bool) (p : Program.t) =
   |> Iter.fold
        (fun p (s, decl) ->
          match decl with
-         | Program.Variable { binding; attrib } when add binding ->
+         | Program_types.Variable { binding; attrib } when add binding ->
```
katrinafyi, line 38: I would prefer that `Program_types` not be used outside of defining Program. It is fully included into Program anyway.
```suggestion
         | Program.Variable { binding; attrib } when add binding ->
```

lib/util/indexed_list.ml
```diff
@@ -0,0 +1,138 @@
+open Containers
+
```
katrinafyi, line 5: Short module-level doc comment explaining motivation and anticipated uses, please :)

lib/util/indexed_list.ml
```diff
@@ -0,0 +1,138 @@
...
+module Make (K : Mtypes.ORD_TYPE) = struct
+  module M = Map.Make (K)
+
+  (* Order stored reverse *)
+
+  type 'a t = { order : K.t list; values : 'a M.t }
```
katrinafyi, line 10: this is an important invariant. we should make it more obvious in the naming.
```suggestion
  type 'a t = { rev_order : K.t list; values : 'a M.t }
```

lib/util/indexed_list.ml
```diff
@@ -0,0 +1,138 @@
...
+    (* we store in reverse order  *)
+    let order = m |> Iter.map fst |> Iter.to_rev_list in
+    (* no duplicate keys  *)
+    assert (Int.equal (M.cardinal values) (List.length order));
+    { values; order }
+
+  let of_list m = List.to_iter m |> of_iter
+
+  (* Get in-order iterator of keys *)
+  let keys { order } = List.to_iter order |> Iter.rev
```
katrinafyi: ocamldoc. here and elsewhere too.
```suggestion
  (** Get in-order iterator of keys *)
  let keys { order } = List.to_iter order |> Iter.rev
```

lib/util/indexed_list.ml
```diff
@@ -0,0 +1,138 @@
...
+      Otherwise it is appended. This operation is the fastest insertion
+      operation (O(1)). *)
+  let append k v { order; values } =
+    if M.mem k values then { order; values = M.add k v values }
+    else { order = k :: order; values = M.add k v values }
+
+  (** [add k v m] replaces k -> v in m, in-place if it is already defined.
+      Otherwise it is appended. This operation is the fastest insertion
+      operation (O(1)). *)
+  let add k v m = append k v m
```
katrinafyi, line 53:
```suggestion
```
As previously, I would not like to have two names for the same thing.

lib/util/indexed_list.ml
```diff
@@ -0,0 +1,138 @@
...
+  let insert_before ~before k v { order; values } =
+    if M.mem k values then { order; values = M.add k v values }
+    else
+      let order = List.rev order in
+      let idx =
+        List.find_index (fun k -> before k (M.find k values)) order
+        |> Option.get_or ~default:(-1)
+      in
+      let order = List.insert_at_idx idx k order |> List.rev in
+      { order; values = M.add k v values }
```
katrinafyi, line 68: The function which would simplify this is Haskell's [break](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Data-List.html#v:break). We don't have that, but we do have something similar in `Extras.span_while_some`. You can either use that, or you can implement something closer to `break` in Extras. If you want.

lib/util/indexed_list.ml
```diff
@@ -0,0 +1,138 @@
...
+  let insert_before_key ~before k v m =
+    insert_before ~before:(fun k _ -> K.equal before k) k v m
+
+  (** Insert value at absolute index. Index is clamped to [0..len-1]. *)
+  let insert_at_index ~before_index k v { order; values } =
+    if M.mem k values then { order; values = M.add k v values }
+    else
+      let idx = List.length order - before_index in
+      let order = List.insert_at_idx idx k order in
+      { order; values = M.add k v values }
```
katrinafyi: CCList docs only say that insert_at_idx clamps in the positive direction. If it's too negative (that is, if the `before_index` is too positive), it will start counting from the end. 

Remove "clamped" comment, or implement clamping, or just assert it's in range (already have prepend/append for those functions).

lib/util/indexed_list.ml
```diff
@@ -0,0 +1,138 @@
...
+    in
+    let pre, post = List.take_drop idx order in
+    let order = List.rev @@ pre @ add_ord @ post in
+    (* update all values  *)
+    let values = List.fold_left (fun acc (k, v) -> M.add k v acc) values news in
+    { order; values }
+
+  let find k { values } = M.find k values
+  let find_opt k { values } = M.find_opt k values
+  let get k { values } = M.find_opt k values
```
katrinafyi, line 108: I would delete one of the `M.find_opt` aliases. Also consider using `_exn` naming.

lib/util/indexed_list.ml
```diff
@@ -0,0 +1,138 @@
...
+
+  (** Out-of-order map *)
+  let map f { order; values } = { order; values = M.map f values }
+
+  (** In-order map with both positional index and key *)
+  let mapi f m =
+    to_iter m |> Iter.mapi (fun ind (k, v) -> (k, f ind k v)) |> of_iter
+
+  let remove k { order; values } =
```
katrinafyi, line 120: could be implemented in terms of `update`.

lib/util/indexed_list.ml
```diff
@@ -0,0 +1,138 @@
...
+    let order = List.sort (Ord.opp compar) order in
+    { order; values }
+
+  (** less efficient sort on values; copies index *)
+  let sort (compar : (K.t * 'a) Ord.t) ({ order; values } : 'a t) =
+    let ord a b =
+      Ord.opp (Ord.map (fun k -> (k, M.find k values)) compar) a b
+    in
+    let order = List.sort ord order in
+    { order; values }
```
katrinafyi, line 148: what is meant by less efficient? both this `sort` and `sort_by_keys` make a copy of the order list.

lib/util/indexed_list.ml
```diff
@@ -0,0 +1,138 @@
...
+    match f (M.find_opt k values) with
+    | Some new_val -> { order; values = M.add k new_val values }
+    | None ->
+        let order = List.remove_one ~eq:K.equal k order in
+        let values = M.remove k values in
+        { order; values }
+
+  (** Sort keys: just List.sort *)
+  let sort_by_keys compar { order; values } =
+    let order = List.sort (Ord.opp compar) order in
```
katrinafyi, line 139: Should this use the intrinsic order on the key type?

lib/transforms/aslp/aslp.ml
```diff
@@ -158,7 +158,7 @@ let add_aarch64_global_declarations ?(include_unused = false) prog =
        (fun prog v ->
          let binding = Aslp_lexpr.to_var v in
          let attrib = Attrib.empty and classification = None in
-         Program.add_decl prog
+         Program.add_decl ~at:`Prepend prog
```
katrinafyi, line 162: i don't think prepend is right. this leads to funny things where the ASLp variables are in exact reverse order:
https://github.com/UQ-PAC/bincaml/blob/b26d4299ab5a80cba13ebb042ca2cbfc26a5fa3d/test/cram/gtirb.t#L30-L41

The originating `Aslp_lexpr.globals` in this iter chain is ordered, so you may wish to collect the undeclared things into a list, then  `insert_list_before`? it would also be nice to have an `~at` which is "before everything which is not a variable/type".

lib/util/indexed_list.ml
```diff
@@ -0,0 +1,138 @@
...
+    in
+    let add_ord =
+      List.filter (fun (id, _) -> not @@ M.mem id values) news |> List.map fst
+    in
+    let pre, post = List.take_drop idx order in
+    let order = List.rev @@ pre @ add_ord @ post in
+    (* update all values  *)
+    let values = List.fold_left (fun acc (k, v) -> M.add k v acc) values news in
+    { order; values }
+
```
katrinafyi, line 100: why is this so complicated? it was okay before getting to this point, but after seeing this function, i think that the `order` reversed list invariant is very burdensome. maybe too burdensome.

have you considered something like [CCFQueue](https://uq-pac.github.io/bincaml/containers-data/containers-data/CCFQueue/index.html)? it might need some manual implementation of list-like utilities, but i think it would be easier (or at least comparable and more clear than) the reversed list + index based code.

test/test_util_indexlist.ml
```diff
@@ -0,0 +1,116 @@
...
+    |}]
+
+open Containers
+
+let%expect_test "invalid from list" =
+  let () = Printexc.record_backtrace false in
+  let m = M.of_list [ (1, 31); (1, 32) ] in
+  pr "remove nonexistent" m;
+  [%expect.unreachable]
+[@@expect.uncaught_exn {| "Assert_failure lib/util/indexed_list.ml:20:4" |}]
```
katrinafyi: assertion expect embeds line number. something like [CCResult.guard_str](https://c-cube.github.io/ocaml-containers/containers/CCResult/index.html#val-guard_str) can be used to try/catch and get the exception string.

katrinafyi (commented):
Good work. The bones are good and the refactoring need for existing uses of declarations seems minimal.

lib/lang/program_types.ml
agle: You need to be able to look up a definition based on a variable (in this case the constructor name), hence the map needs to be bijective and one entry in the map shouldn't declare 10 entries.

katrinafyi: Roger roger.

lib/transforms/gamma_vars.ml
agle: This doesn't work for whatever reason, in other places I've used a type hint `Program.declaration`.

lib/lang/program.ml
```diff
@@ -25,35 +21,23 @@ let show_stmt =
...
-    }
-  | Procedure of { definition : proc }
+let decl_is_typ = function Type _ -> true | _ -> false
+let decl_is_func = function Function _ -> true | _ -> false
+let decl_is_var = function Variable _ -> true | _ -> false
+let decl_is_proc = function Procedure _ -> true | _ -> false
+let types_before _ k = not (decl_is_typ k)
+let vars_before _ k = decl_is_proc k || decl_is_func k || decl_is_var k
+let procs_before _ k = decl_is_proc k
+let funcs_before _ k = decl_is_func k
```
katrinafyi: These names have no meaning without knowing the context. i think these bindings are not necessary and `fun _ -> decl_is_proc` or similar, at the `DeclsList.insert_before` call sites, would be fine. alternatively, move these into `DeclsList` or provide partially-applied versions of `DeclsList.insert_before` inside DeclsList.

lib/util/indexed_list.ml
agle: It's not complicated. It reverses the order, finds the index to insert at, splits on that index, puts the new list in the middle, then reverses again. I have not considered CCFQueue, changing to it would amount to a rewrite of this PR, and can be done in a later pr.

lib/util/indexed_list.ml
agle: That would just make this function less useful, it is intended to be analogous to List.sort.

lib/util/indexed_list.ml
agle: It has to look up the map for the values

lib/util/indexed_list.ml
agle: This exists to meet user expectations of satisfying the Map interface.

lib/util/indexed_list.ml
```diff
@@ -0,0 +1,138 @@
...
+  let singleton k v = { order = [ k ]; values = M.singleton k v }
+
+  let of_iter m =
+    let m = Iter.persistent m in
+    let values = M.of_iter m in
+    (* we store in reverse order  *)
+    let order = m |> Iter.map fst |> Iter.to_rev_list in
+    (* no duplicate keys  *)
+    assert (Int.equal (M.cardinal values) (List.length order));
+    { values; order }
```
katrinafyi, line 23: `.persistent` is quite the burden on memory. This could be simplified to just use `append`. There is some extra bookkeeping to make sure they're distinct.
```ocaml
  let of_iter m =
    let iter_len = ref 0 in
    let m =
      m
      |> Iter.map (Fun.tap (fun _ -> iter_len := !iter_len + 1))
      |> Iter.fold (fun x (k, v) -> append k v x) empty
    in
    assert (M.cardinal m.values = !iter_len);
    m
```

katrinafyi (commented):
Sorry, I had these two thoughts as well but I forgot to type them up.

lib/transforms/aslp/aslp.ml
agle: The order is not semantic I don't really care if it is reversed. Reversing the order is exactly what you expect from sequentially prepending. If it matters, a fold_right instead of left would insert in the correct order.

lib/util/indexed_list.ml
katrinafyi: I think it's pretty burdensome, but roger roger.

lib/util/indexed_list.ml
agle: This function is fundamentally O(n) space complexity. I would guess its a more significant performance gain to use the list and maps' specialised constructors from sequences than a iterated insertion.

lib/transforms/aslp/aslp.ml
katrinafyi: I meant "I don't think prepend is the right mode to use here". But understood.

lib/util/indexed_list.ml
katrinafyi: Understood.

lib/util/indexed_list.ml
katrinafyi: That's marginal and, in fact, expected given the functionality. I wouldn't feel the need to mention that in the doc comment. But up to you if change is needed.

lib/util/indexed_list.ml
katrinafyi: Then adjust doc comment to reference the function which it is equivalent to.

lib/util/indexed_list.ml
katrinafyi: I'm not saying it would be less asymptotically, but the persistent iter is a third object in addition to the needed values + order. But acknowledged.

lib/lang/program.ml
agle: Not part of this pr, I belive previously add_decl would exn if the decl already existed, but that the responsibility for ID generation has been moved since. It can be deleted in a separate pr, as it requries a project-wide renaming.

lib/util/indexed_list.ml
agle: This would make it more complicated in fact; I don't need the two lists, only need the index in order to `insert_at_index` on the key list. Haskell's break is just our `take_drop_while` with the test negated?...which insert_list_before does already use. It would still require reversing the list.

lib/util/indexed_list.ml
agle: Again this is to satisfy the conventional interfaces for both associative and sequential datastructures.

lib/util/indexed_list.ml
agle: ~~No need to unneccessarily allocate options for a slightly different implementation of a one-line function.~~ 

update is implemented in terms of remove.

lib/util/indexed_list.ml
agle: The comparison is log(n) instead of constant and is the reason I made sort on index a separate function.

lib/util/indexed_list.ml
agle: CCFQueue doesn't implement insertion in the middle or something like `take_drop_while` so while appending would be more efficient it would only make this function more complex.

lib/lang/program.ml
agle: Pretty functions must be parseable

katrinafyi: With very very few exceptions, every type should be printable somehow. If not here, you can add a `deriving show` to the declarations type.

lib/lang/program.ml
agle: I did this cause they're already hidden by the mli file but sure they can go in the function instead.

lib/util/indexed_list.ml
katrinafyi: I'm not disputing that it is slower, nor am I against the separate functions. I just think "less efficient sort on values; copies index" is way too terse and only invites more questions.

katrinafyi (approved):
?

PR #237, opened 2026-08-14
Add some ops missing from aslp generator

agle:
- `cvt_bool_bv`
- `replicate_bits`
- fix missing cases of `\`NEQ` in Ops.ml
- try not emit assert(or())


lib/transforms/aslp/bincaml_ibi_make.ml
```diff
@@ -413,7 +428,9 @@ struct
   (** [f_gen_replicate_bits operand_width num_replications operand
        num_replications] *)
   let f_gen_replicate_bits : bigint -> bigint -> expr -> expr -> expr =
-   fun _ -> failwith "f_gen_replicate_bits"
+   fun targ0 targ1 opr nr ->
+    try_const_unsigned_value nr |> Z.to_int |> fun repeats ->
```
katrinafyi: should be no need for `try_const_unsigned_value`; it should be in `targ1` which should be a bigint equal to `nr`'s value.

lib/gtirb/gfir_to_bincaml.ml
```diff
@@ -31,7 +31,9 @@ let add_proxy_block ?(attrib = StringMap.empty) succ_addr (proc, blockmap) uuid
     succ_addr uuid
     |> Iter.map (fun addr -> addr_equal_expr addr)
     |> Iter.to_list
-    |> Expr.BasilExpr.applyintrin ~op:`OR
+    |> function
+    | [] -> Expr.BasilExpr.boolconst true
+    | args -> Expr.BasilExpr.applyintrin ~op:`OR args
```
katrinafyi, line 36: Did this cause a problem somewhere downstream? Generally, it would be nice to handle 0-ary associative operations sensibly.

katrinafyi (approved):
LGTM modulo comments. Thanks :)

lib/gtirb/gfir_to_bincaml.ml
agle: Generally I just give them semantics, SMTLib and boogie don't like it tho so the expr simplifier converts them to unit. Replacing them in the constructor would be unexpected I think.

katrinafyi: If it is to be replaced in this one place, you could imagine that it would also need similar replacements in other random places. Regardless of how it's done, it would be nice to centralise this somewhere to avoid needing ad-hoc patches. But not necessarily in this PR, sure.

lib/transforms/aslp/bincaml_ibi_make.ml
agle: Are you sure its not just the bitwidth of the result?

lib/gtirb/gfir_to_bincaml.ml
agle: It is centralised in Expr_simp :)

lib/transforms/aslp/bincaml_ibi_make.ml
katrinafyi: Reasonably. Did you find it to not be the case?

agle: No but I can't tell from this example

katrinafyi: Best citation I can show.. https://github.com/UQ-PAC/aslp/blob/92e4fd949402056e16e3747e66922060be57ecc8/offlineASL/template_offline_utils.ml#L260-L261

agle: k

PR #238, opened 2026-08-18
Use CCFQueue for Indexed_list

katrinafyi:
This changes the Indexed_list data structure to use a [CCFQueue](https://uq-pac.github.io/bincaml/containers-data/containers-data/CCFQueue/index.html) because it has faster append on both ends (amortised O(1)) and not worse insertion in the middle (linear). More importantly, it also means that the order does not have to be stored in a reversed list order which uses index-based slicing.

Along the way, I have added printers for debugging and odoc comments that clarify some edge cases.

I have also found and fixed some bugs where `insert_at_index` did not do what the comment said for out of bounds indices (now simply asserted against) and where `update` would lead to an incomplete order when the `f` function returns Some for a key which was previously None.

Related https://github.com/UQ-PAC/bincaml/pull/236 https://github.com/UQ-PAC/bincaml/issues/235

### DIFF
--- a/lib/util/indexed_list.ml
+++ b/lib/util/indexed_list.ml
@@ -153,9 +153,18 @@ module Make (K : Mtypes.ORD_TYPE) = struct
 
   (** {2 Conversions and iterators} *)
 
-  (** Builds a {!t} from the given pairs. Repeated keys use the order of the
-      first occurence and the value of the last occurence. *)
-  let of_iter m = Iter.fold (fun m (k, v) -> append k v m) empty m
+  (** Builds a {!t} from the given pairs. Keys must be unique. *)
+  let of_iter m =
+    let n = ref 0 in
+    let m =
+      Iter.fold
+        (fun m (k, v) ->
+          n := !n + 1;
+          append k v m)
+        empty m
+    in
+    if !n <> length m then invalid_arg "of_iter: had duplicated keys";
+    m
 
   let of_list m = List.to_iter m |> of_iter
 
--- a/lib/util/indexed_list.ml
+++ b/lib/util/indexed_list.ml
@@ -20,6 +20,7 @@ module Make (K : Mtypes.ORD_TYPE) = struct
   open struct
     module S = Set.Make (K)
     module M = Map.Make (K)
+
     let pp_k = Format.of_to_string K.show
   end
 
@@ -106,7 +107,7 @@ module Make (K : Mtypes.ORD_TYPE) = struct
     in
 
     let before, after =
-      FQ.take_front_while (fun k -> before k (M.find k values)) order
+      FQ.take_front_while (fun k -> not (before k (M.find k values))) order
     in
     let order = FQ.cons_l before (FQ.cons_l new_keys after) in
     { order; values = M.add_list values news }
@@ -146,7 +147,7 @@ module Make (K : Mtypes.ORD_TYPE) = struct
   let get_exn k { values } = M.find k values
 
   let update k f m =
-    match f (find k m) with
+    match f (find_opt k m) with
     | Some new_val -> add k new_val m
     | None -> remove k m
 
--- a/lib/util/indexed_list.ml
+++ b/lib/util/indexed_list.ml
@@ -37,8 +37,7 @@ module Make (K : Mtypes.ORD_TYPE) = struct
 
   let empty = { order = FQ.empty; values = M.empty }
   let singleton k v = { order = FQ.singleton k; values = M.singleton k v }
-  let cardinal m = M.cardinal m.values
-  let length m = cardinal m
+  let size m = FQ.size m.order
 
   (** {2 Inserting, updating, and removing} *)
 
@@ -163,7 +162,7 @@ module Make (K : Mtypes.ORD_TYPE) = struct
           append k v m)
         empty m
     in
-    if !n <> length m then invalid_arg "of_iter: had duplicated keys";
+    if !n <> size m then invalid_arg "of_iter: had duplicated keys";
     m
 
   let of_list m = List.to_iter m |> of_iter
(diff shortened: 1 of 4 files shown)


lib/util/indexed_list.ml
```diff
@@ -1,104 +1,140 @@
...
 
-  (** Get in-order list of pairs of keys and values *)
-  let to_list m = m |> to_iter |> Iter.to_list
+      [order] must be a permutation of the keys of [values]. *)
 
-  (** Values iterator in-order *)
-  let values m = to_iter m |> Iter.map snd
+  let empty = { order = FQ.empty; values = M.empty }
+  let singleton k v = { order = FQ.singleton k; values = M.singleton k v }
+  let cardinal m = M.cardinal m.values
```
ncough: Last I checked `M.cardinal` was O(n). Probably best to use `FQ.size`? Seems they cache it.

lib/util/indexed_list.ml
```diff
@@ -1,104 +1,140 @@
...
 
-  (** Values iterator out-of-order *)
-  let values_nondet { values } = M.values values
+  (** {2 Inserting, updating, and removing} *)
 
-  (** [append k v m] replaces k -> v in m, in-place if it is already defined.
-      Otherwise it is appended. This operation is the fastest insertion
-      operation (O(1)). *)
+  (** [append k v m] replaces [k -> v] in m, in-place if it is already defined.
+      Otherwise it is appended. This operation is fast (O(1)). *)
```
ncough, line 45: I'm not sure `append`, `add` and `prepend` are really O(1), given the map operations. Not that they have to be, just possibly misleading.

lib/util/indexed_list.ml
```diff
@@ -1,104 +1,140 @@
...
-  let append_list news m = List.fold_left (fun acc (k, v) -> add k v acc) m news
+  (** Inserts the given pairs, as a block, before the first element satisfying
+      the given [before] predicate.
+
+      If any new keys were already in the map, those keys will be updated
+      in-place. *)
+  let insert_list_before ~before news { values; order } =
+    (* ignore pre-existing keys and only use leftmost position of each new key. *)
+    let _, new_keys =
+      let old_keys_set = S.of_iter (Iter.map fst (M.to_iter values)) in
```
ncough: `M.keys` rather than the `map fst`? Pretty minor.

lib/util/indexed_list.ml
```diff
@@ -110,40 +146,64 @@ module Make (K : Mtypes.ORD_TYPE) = struct
...
+  let of_iter m =
+    let n = ref 0 in
+    let m =
+      Iter.fold
+        (fun m (k, v) ->
+          n := !n + 1;
+          append k v m)
+        empty m
+    in
+    if !n <> length m then invalid_arg "of_iter: had duplicated keys";
```
ncough: I guess this legitimately needs the map size.

lib/util/indexed_list.ml
```diff
@@ -1,104 +1,140 @@
...
       { order; values = M.add k v values }
 
-  (** Relatively efficient: cons *)
-  let append_list news m = List.fold_left (fun acc (k, v) -> add k v acc) m news
+  (** Inserts the given pairs, as a block, before the first element satisfying
+      the given [before] predicate.
+
+      If any new keys were already in the map, those keys will be updated
+      in-place. *)
+  let insert_list_before ~before news { values; order } =
```
ncough, line 99: Pretty minor, but can you avoid the set and just build the final map while deduplicating? Maybe the current is easier to understand.

lib/util/indexed_list.ml
```diff
@@ -110,40 +146,64 @@ module Make (K : Mtypes.ORD_TYPE) = struct
...
+
+  (** {2 Mapping and sorting} *)
+
   (** Out-of-order map *)
   let map f { order; values } = { order; values = M.map f values }
 
   (** In-order map with both positional index and key *)
   let mapi f m =
-    to_iter m |> Iter.mapi (fun ind (k, v) -> (k, f ind k v)) |> of_iter
+    to_iter m |> Iter.mapi (fun i (k, v) -> (k, f i k v)) |> of_iter
```
ncough: Seems a little redundant to check for collisions and rebuild the order here. Could this iterate order and update map accordingly?

ncough (commented):
Pretty nifty stuff. Comments are mostly performance related, so feel free to dismiss if they aren't worth it.

lib/util/indexed_list.ml
katrinafyi: rewritten to use a rewritten `update` function and avoid repeated lookups into the map.

lib/util/indexed_list.ml
katrinafyi: I decided not to change insert_list_before because in the case where `news` has duplicate keys, repeatedly prepending will use the rightmost position of that key rather than the leftmost, which might be counterintuitive. A corner case, for sure, but I'm just going to leave
it for now.

Another possibility I looked at was using `of_list` on `news` but inserting a Indexed_list at a certain position is just as hard.

lib/util/indexed_list.ml
katrinafyi: It is an invariant that map size should always equal the FQ size.

lib/util/indexed_list.ml
katrinafyi: Removed explicit claim.

lib/util/indexed_list.ml
katrinafyi: Good spot. I never checked.

katrinafyi (commented):
Thanks. Btw the rules say that you (or someone else) will have to click the green button before it can be merged.

ncough (approved):

PR #239, opened 2026-08-18
    ci: run cached `opam` directly in GHA and add `nix` jobs

katrinafyi:
Supersedes https://github.com/UQ-PAC/bincaml/pull/230

This is kind of tricky. We cannot go all-in on Opam because we have forks of bnfc and those cannot be easily installed by depext. We cannot go all-in on Nix because this an Opam project firstly. Mixing the two and having a fast Nix fall back to a slow opam was also unsatisfying because the YML turned really bad.

In the end, I just have made `nix` and `opam` jobs and they all race and run in parallel. The Opam jobs still use Nix in a limited capacity to install some of the bnfc and SMT programs. 

I expect that the PR required gate should be set to the opam jobs only.

Both opam and nix approaches are cached, so they have converged to both needing about 1-1.5 minute of setup before doing actual work. This is okay, I think. The Nix is using an off-the-shelf action: https://github.com/nix-community/cache-nix-action, whereas the Opam is using a custom cache action that lives in `./.github/actions/setup-bincaml`. I am cautiously optimisic that the custom opam switch caching will work well. I have tried to add a monthly refresh so it should pull in new opam updates at least monthly, but that is yet to be tested.

In hindsight, it does seem like a docker container was the best of both worlds, if only the random crash could be fixed. Maybe that is worth looking in to if you don't want this PR. It does not have a great story around automatic updates, though.


in-nix
mira-alford, line 1: Neat!

mira-alford (approved):

PR #247, opened 2026-08-24
Mira ssa

mira-alford:
My attempt at SSA.
Work in progress!

Goals (in rough order):
- [X] SSA Construction
- [X] SSA Destruction

Leaving reconstruction for a future PR.


lib/transforms/may_read_uninit.ml
```diff
@@ -1,43 +1,54 @@
...
-    | a, b when equal a b -> true
-    | Bot, _ | _, ReadUninit -> true
-    | _, Bot | ReadUninit, _ -> false
-    | _ -> false
+    | Bot, _ -> true
+    | _, Bot -> false
+    | Val (s1, v1), Val (s2, v2) -> leq_state s1 s2 && leq_val v1 v2
+
+  let show_state s = match s with Init -> "I" | Uninit -> "U"
+  let show_val v = match v with Happy -> ":)" | Sad -> ":("
```
b-paul: :3

lib/transforms/may_read_uninit.ml
```diff
@@ -52,30 +63,47 @@ module ReadUninitAnalysis = struct
...
   let write_var st v =
     match read st v with
-    | ReadUninit -> ReadUninit.ReadUninit
-    | Write -> Write
-    | Bot -> Write
+    | Bot -> ReadUninit.Bot
+    | Val (Uninit, v) -> Val (Init, v)
+    | o -> o
 
   let read_uninit_vars st =
```
b-paul, line 76: i think this is confusingly named, since it actually gets the variables that were read when uninitialised, it doesn't read all of the uninit vars (how i interpreted it at first)

lib/transforms/may_read_uninit.ml
```diff
@@ -52,30 +63,47 @@ module ReadUninitAnalysis = struct
...
-    Procedure.formal_in_params p
-    |> Common.StringMap.values
-    |> Iter.fold (fun acc v -> update v ReadUninit.Write acc) bottom
+  let init ?(vertex=None) (p:Program.proc) =
+    let args =
+      Procedure.formal_in_params p
+      |> Common.StringMap.values
+      |> Iter.fold
+           (fun acc v -> update v (ReadUninit.Val (Init, Happy)) acc)
+           bottom
```
b-paul, line 92: I think this should only mark args as init at entry instead of globally on the whole graph. I'd think we want the check for being Vert.Entry at the start before computing args or locals, and else return bottom

lib/transforms/drop_unused.ml
```diff
@@ -0,0 +1,34 @@
...
+    Procedure.fold_blocks_topo_fwd
+      (fun acc id bl ->
+        Iter.append (Block.read_vars_iter bl) (Block.assigned_vars_iter bl)
+        |> Iter.fold (fun acc i -> VarSet.add i acc) acc)
+      VarSet.empty p
+  in
+  Var.Decls.filter_map_inplace
+    (fun _ v -> if VarSet.mem v used then Some v else None)
+    (Procedure.local_decls p);
+  VarSet.filter Var.is_global used
```
b-paul, line 17: grumble grumble the name of this function should describe it's return value being the used variables ! ! it's confusing that it both mutates and collects information but maybe a doc comment would help

lib/transforms/lambda_lifting.ml
```diff
@@ -0,0 +1,384 @@
+(** Naive parameter form. *)
```
b-paul, line 1: Is parameter form lambda lifting?

lib/passes.ml
```diff
@@ -208,7 +224,7 @@ module PassManager = struct
   let sssa =
     {
       name = "simple-ssa";
-      apply = Proc Transforms.Ssa.ssa;
+      apply = Transforms.Ssa.(Prog (ssa_prog ~skipping:Skip.full));
```
b-paul, line 227: why not put the Transforms.Ssa after Prog 😭

b-paul (commented):
I didn't look at ssa.ml so this review is incomplete but meow

mira-alford:
Still need to pull reaching defs out into a seperate PR for now but otherwise it is approaching completeness. Just a test failing.

lib/transforms/ssa.ml
```diff
@@ -111,595 +71,344 @@ let check_ssa ~skip_observable ~skip_maps proc =
...
-  let globals_to_param_lift =
-    Program.procs p
+                    (src, stmt)))
+            |> Iter.fold
+                 (fun acc (id, stmt) ->
+                   Procedure.modify_block acc id (fun b ->
+                       Block.append_stmts b [ stmt ]))
+                 acc)
+          procedure procedure
+      in
```
b-paul, line 106: this is so nice!! 😍

lib/transforms/ssa.ml
```diff
@@ -111,595 +71,344 @@ let check_ssa ~skip_observable ~skip_maps proc =
...
+              | _ -> raise Not_found
+            in
+            let block = { block with phis = phi :: block.phis } in
+            graph := G.remove_edge !graph (Begin id) (End id);
+            graph := G.add_edge_e !graph (Begin id, Block block, End id)
+        | _ -> ())
+    done;
+    !graph
+
+  (* Helper to modify a block at id *)
```
b-paul, line 193: should this be a doc comment ?

lib/transforms/ssa.ml
```diff
@@ -111,595 +71,344 @@ let check_ssa ~skip_observable ~skip_maps proc =
...
+              match G.find_edge !graph (Vert.Begin id) (Vert.End id) with
+              | _, Block b, _ -> b
+              | _ -> raise Not_found
+            in
+            let block = { block with phis = phi :: block.phis } in
+            graph := G.remove_edge !graph (Begin id) (End id);
+            graph := G.add_edge_e !graph (Begin id, Block block, End id)
+        | _ -> ())
+    done;
+    !graph
```
b-paul, line 191: This is awesome and really nice to read but i wonder if perhaps this could be made more functional by computing like a dominance frontier closure of the defs, (fixpoint of the map S |-> S union all dominance frontiers of vertices in S) then adding phis to all of the vertices in that set (or perhaps deduped iter !)... but it's definitely fine as it is

lib/transforms/ssa.ml
```diff
@@ -111,595 +71,344 @@ let check_ssa ~skip_observable ~skip_maps proc =
...
+               phi.rhs
+               |> List.map (function
+                 | id, rvar when ID.equal id par_id ->
+                     (id, rename_rvar (End par_id) rvar)
+                 | o -> o)
+             in
+             { phi with rhs }))
+      Fun.id block
+
+  (* Update a the reaching def of var by climbing up the reaching
```
b-paul, line 278: a the

lib/transforms/ssa.ml
```diff
@@ -111,595 +71,344 @@ let check_ssa ~skip_observable ~skip_maps proc =
...
-              rhs
-              |> List.map (fun (a, b) ->
-                  (a, VarMap.get_or ~default:b b (get_st_pred a))) )
-            acc)
-        VarMap.empty cur_phis
+          List.fold_left
+            (fun acc v ->
+              let acc = G.remove_edge acc v Return in
+              G.add_edge acc v (Begin rid))
+            g returns)
```
b-paul, line 326: tiny nitpick but could you move this function to outside of the renaming code ?

lib/transforms/ssa.ml
```diff
@@ -111,595 +71,344 @@ let check_ssa ~skip_observable ~skip_maps proc =
...
-            |> List.fold_left
-                 (fun phim (block, rn) ->
-                   let rn = VarMap.filter (fun v _ -> VarSet.mem v lives2) rn in
-                   VarMap.merge_safe ~f:(merge_phi block) phim rn)
-                 cur_phis
+        (* Add definition to defs. *)
+        defs := VarMap.add var' vert !defs;
+        continue k var'
+    | effect AllowRename vert, k ->
+        continue k (not @@ Vert.equal vert (Begin rid))
```
b-paul, line 368: Wow effects really make this imperative algorithm so clean in ocaml... we should have more effects in the codebase! I do wonder though if there's a nicer more functional way to do renaming since this was kinda hard to follow before i got that there was lots of reaching def state relative to the traversal order but that's probably out of scope of this pr

lib/transforms/ssa.ml
```diff
@@ -111,595 +71,344 @@ let check_ssa ~skip_observable ~skip_maps proc =
...
+        ~stmts:
+          (Procedure.formal_out_params procedure
+          |> StringMap.values
+          |> Iter.map (fun v ->
+              Stmt.Instr_Assign
+                {
+                  al = [ (v, Expr.BasilExpr.rvar v) ];
+                  attrib = StringMap.empty;
+                })
+          |> Iter.to_list)
```
b-paul, line 312: This could probably be a simultaneous assign which would make generated return blocks a little smaller code size wise.

b-paul (approved):
This is great!! the ssa algorithm is very readable :) my comments are only minor nitpicks so after those changes this is probably good to merge given the tests seem to mirror old ssa functionality.

PR #250, opened 2026-08-26
Ide analysis documentation

b-paul:
Here are some docs on how to write an IDE analysis.

I'm not much of an expert in formatting and whatnot so if you have any comments please share ! I hope the content is at least somewhat understandable :3

### DIFF
--- a/docs/ide.mld
+++ b/docs/ide.mld
@@ -64,44 +64,5 @@ For SSA reasons, identity edges on variables or the sink do not need to be given
 - [init_param_values] specifies the initial value lattice values for in/out params of a procedure.
 Like with [relevant_params], you probably only care about in params if forwards and out params if backwards.
 
-{[
-module type IDESSIDomain = sig
-  (* Edge functions *)
-  include Lattice
 
-  module DL : sig
-    type t = Lambda | Label of Var.t
-    [@@deriving eq, ord, show { with_path = false }]
-  end
-
-  val direction : [ `Forwards | `Backwards ]
-  (** The direction this analysis should be performed in *)
-
-  module Value : Lattice
-  (** The underlying lattice the edge functions operate on *)
-
-  val identity : t
-  (** identity edge function *)
-
-  val top : t
-  (** top edge function *)
-
-  val compose : t -> t -> t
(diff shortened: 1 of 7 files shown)


lib/analysis/wrapped_intervals.ml
```diff
@@ -720,7 +720,7 @@ module WrappedIntervalsValueAbstraction = struct
     | _ -> top
 
   let eval_binary op (a, ta) (b, tb) rt =
-    match (ta, ta) with
+    match (ta, tb) with
```
katrinafyi, line 723: so true 😨

docs/ide.mld
```diff
@@ -0,0 +1,107 @@
...
+If the analysis is forwards you probably want to set in params and if it is backwards you probably want to set out params.
+- [transfer_call] specifies the edges from a callee variable/sink into input parameters of a caller function based on the arguments in a call statement.
+- [transfer] specifies the edges out of a variable/sink specified by a statement.
+For SSA reasons, identity edges on variables or the sink do not need to be given, unlike in non-SSA IDE analyses.
+- [transfer_phi] specifies edges between a phi node.
+- [init_param_values] specifies the initial value lattice values for in/out params of a procedure.
+Like with [relevant_params], you probably only care about in params if forwards and out params if backwards.
+
+{[
+module type IDESSIDomain = sig
```
katrinafyi: I am wary of pasting large signatures into docs as it might easily fall out of date. What do you think?

A link to the docs should usually be enough. You can also link to the "Source" view if you want that to be visible.

docs/ide.mld
```diff
@@ -0,0 +1,107 @@
+{0 Writing IDE analyses}
+
+{1 Introduction}
+
```
katrinafyi, line 4: do you think it needs citation to the paper, or is that already in the code comments elsewhere? any other blog posts or things like that on this topic?

katrinafyi (approved):
LGTM! It looks well written and easy to follow. Just small questions and then we can merge.

docs/ide.mld
b-paul: you're so right a link makes way more sense

PR #251, opened 2026-08-27
Prototype for Rely-Guarantee Generation

j-tobler:
Main contributions:
- Implemented some basic logic for the generation of rely-guarantee conditions using interference domains.
- Fixed some bugs in LatticeMap.
- Slightly altered the design of WrappedIntervals.

The InterferenceStateDomain module extends Intra_analysis.IntraDomain with some functions required to be a compatible underlying state domain for rely-guarantee generation.
An InterferenceDomain is an abstract domain for rely-guarantee conditions. One of these - the ConditionalWritesDomain - has been implemented.
The RelyGuaranteeGenerator contains a set of functions which together derive rely-guarantee conditions in the form of a chosen InterferenceDomain, given an underlying InterferenceStateDomain. This involves repeatedly deriving reachable states and guarantee conditions for each thread until a fixpoint is reached.

WrappedIntervals used to initialise to bottom, which means forward intra analyses using the domain would begin with bottom as a precondition / initial state. I'm pretty sure I might be missing something here, but to make RG-gen work properly, I've changed it so that it initialises to top. Please let me know if I've missed some way of specifying preconditions for analyses.

Some other possible issues with the current implementation:
- I've left some toggle-able debug print statements in the code. These are useful to me but let me know if they can't go in main.
- I wasn't sure how to implement transfer_phi but this is quite a rough prototype and I'm sure this will be added to the next version.


lib/analysis/rg_gen.ml
```diff
@@ -0,0 +1,392 @@
...
+
+open Bincaml_util.Common
+open Lattice_types
+open Lang
+open Expr
+
+module Debug = struct
+  let enabled = false
+  let print s = if enabled then print_endline s
+end
```
agle, line 11: Use `Logs.debug (fun m -> m "string")`, this will save it from computing the string when debugging is disabled.

lib/analysis/rg_gen.ml
```diff
@@ -0,0 +1,392 @@
...
+    where each assignment may simultaneously assign a set of global variables.
+    Note that this means ignoring any local variables that may also be updated
+    in the assignment. 2.d. From those concrete interferences, use the
+    I.transitions function to derive a guarantee condition (as an I.t) that
+    over-approximates the set of state transitions that may be induced by their
+    execution. 3. Update each thread's guarantee condition in the map to the new
+    one derived from step 2. 4. Repeat steps 2-3 until a fixpoint is reached
+    over the derived guarantee conditions. 5. The generated guarantee conditions
+    are given by the map. The generated rely conditions for each thread are
+    given by the join over all environment guarantee conditions.
```
agle, line 276: You maybe formatted these steps in lists and it got messed up by the autoformatter; consider using list tags in the odoc syntax: https://ocaml.github.io/odoc/odoc/odoc_for_authors.html#lists

lib/analysis/rg_gen.ml
```diff
@@ -0,0 +1,392 @@
...
+      include D
+
+      let name = D.name ^ " Stabilised"
+
+      (** Stabilise the pre-state before transferring. I believe this may be
+          soundly modified such that stabilisation only occurs before a global
+          read or write. *)
+      let transfer t = D.transfer @@ I.stabilise rely t
+      (* todo: what about transfer_phi? *)
+    end)
```
agle, line 315: As far as I can tell this can just be a regular module instead of a first class module? 

```ocaml
module StableD = struct 
   include D
   let name = D.name ^ "stable" 
   let transfer t = D.transfer @@ I.stabilise rely t
end
module Analysis = Intra_analysis.Forwards (StableD)
```

lib/analysis/rg_gen.ml
```diff
@@ -0,0 +1,392 @@
...
+    | [] -> [ [] ]
+    | x :: xs ->
+        let sets_without_x = powerset xs in
+        let sets_with_x = List.map (fun sublst -> x :: sublst) sets_without_x in
+        sets_without_x @ sets_with_x
+
+  (** General function for deriving a fixpoint. *)
+  let rec fixpoint equal f x =
+    let y = f x in
+    if equal y x then y else fixpoint equal f y
```
agle, line 26: We have the same function defined in `Lang.Algsimp.to_steady`; this could be defined in `Bincaml_util.Extras` and shared

Also note there is https://uq-pac.github.io/bincaml/fix/index.html#memoization-and-tabulation for memoisation and tabulation of  monotonic recursive functions.

lib/analysis/rg_gen.ml
```diff
@@ -0,0 +1,392 @@
...
+    For two variable sets x and y mapping to write-conditions P and Q, we
+    maintain the invariant that (x U y) maps to a write-condition stronger than
+    P /\ Q. Thus P and Q are sufficient over-approximations of the states under
+    which x and y can be written to respectively, so we avoid having to look
+    through the write-conditions of their supersets when determining the
+    conditions under which they can change. Note that (x U y) may map to a
+    strictly stronger write-condition than P /\ Q, such as in the case when
+    either x or y can change individually but never in the same execution trace
+    (i.e. "at the same time"). *)
+module ConditionalWritesDomain (D : InterferenceStateDomain) = struct
```
agle, line 159: Probably a good idea the paper you wrote about this here! https://link.springer.com/chapter/10.1007/978-3-032-26204-2_20

lib/analysis/rg_gen.ml
```diff
@@ -0,0 +1,392 @@
...
+            ConcInt.{ pre = stable_pre; assignments = global_assigns }
+            :: interferences
+      | _ -> interferences
+    in
+    { state = post; rely; interferences = new_interferences }
+
+  let derive_guar rely proc =
+    let (module StableD) = stable_d rely in
+    let module Analysis = Intra_analysis.Forwards (StableD) in
+    let analysis_results = Analysis.analyse proc in
```
agle, line 343: This (`analysis_results`) doesn't seem to depend on rely, and I don't think proc is changed throughout the analysis, so this analysis result could be stored rather than re-computed for every call to `rederive_guar`?

lib/analysis/rg_gen.ml
```diff
@@ -0,0 +1,392 @@
...
+    maintain the invariant that (x U y) maps to a write-condition stronger than
+    P /\ Q. Thus P and Q are sufficient over-approximations of the states under
+    which x and y can be written to respectively, so we avoid having to look
+    through the write-conditions of their supersets when determining the
+    conditions under which they can change. Note that (x U y) may map to a
+    strictly stronger write-condition than P /\ Q, such as in the case when
+    either x or y can change individually but never in the same execution trace
+    (i.e. "at the same time"). *)
+module ConditionalWritesDomain (D : InterferenceStateDomain) = struct
+  module VarSetMap = Map.Make (VarSet)
```
agle, line 160: Maybe for the next iteration: consider using Patricia tree map and hash-consed var set: https://uq-pac.github.io/bincaml/patricia-tree/index.html, if the state domain `D` is hashable. (`Var` is already hashable).

This might make `union` more efficient by making comparison of equal subtrees constant-time (see [`idempotent_union`](https://uq-pac.github.io/bincaml/patricia-tree/patricia-tree/PatriciaTree/MakeMap/index.html#val-idempotent_union)).

lib/analysis/wrapped_intervals.ml
```diff
@@ -785,7 +787,7 @@ module Domain = struct
     let vs = Lang.Procedure.formal_in_params p |> StringMap.values in
     vs
     |> Iter.map (fun v -> (v, top_val))
-    |> Iter.fold (fun m (v, d) -> update v d m) bottom
+    |> Iter.fold (fun m (v, d) -> update v d m) top
```
ncough: I think the original `bottom` is correct here, as it's used to represent variables that have not yet been assigned a value. I suspect you want is to instead initialise globals/shared variables to `top` somehow, rather than changing this.

lib/analysis/rg_gen.ml
```diff
@@ -0,0 +1,392 @@
...
+    let result = bot_binop WrappedIntervalsLattice.meet t1 t2 in
+    Debug.print @@ "Result " ^ show result ^ "\n";
+    result
+
+  let havoc t var_set =
+    Debug.print @@ "Havocing "
+    ^ VarSet.to_string ~start:"{" ~stop:"}" ~sep:"," Var.show var_set
+    ^ " in " ^ show t;
+    let result =
+      if Domain.contains_bot t then t
```
ncough: I think you'd want to guard this by a `bot` global/shared variable.

lib/analysis/rg_gen.ml
```diff
@@ -0,0 +1,392 @@
...
+        VarSet.fold
+          (fun var acc -> update var WrappedIntervalsLattice.top acc)
+          var_set t
+    in
+    Debug.print @@ "Result " ^ show result ^ "\n";
+    result
+
+  let filter t exp =
+    transfer t
+    @@ Stmt.Instr_Assume { attrib = Attrib.empty; body = exp; branch = false }
```
ncough: I think perhaps I see the issue now. Does this represent an infeasible guard by reducing variable abstractions to `bottom`?

lib/analysis/rg_gen.ml
```diff
@@ -0,0 +1,392 @@
...
+        let assign_v' =
+          D.transfer pre
+            (Stmt.Instr_Assign { attrib = Attrib.empty; al = [ (v', e) ] })
+        in
+        (* create the expression v' != v *)
+        let v_not_eq_v' = BasilExpr.binexp ~op:`NEQ v_exp v'_exp in
+        (* apply filter v' != v to the state resulting from v' := e *)
+        let filtered = D.filter assign_v' v_not_eq_v' in
+        (* if the result is bot, then v' == v must always hold after v' := e, meaning v := e doesn't change v *)
+        not @@ D.equal filtered D.bottom
```
ncough: I'm not sure this can trigger, as the most the above can do is fix some variable to `bottom`, not the full state.

lib/analysis/lattice_collections.ml
```diff
@@ -213,6 +223,8 @@ module LatticeMap (K : MapKey) (V : TopLattice) = struct
         val singleton : K.t -> V.t -> t
         val mapi : (K.t -> V.t -> V.t) -> t -> t
         val fold : (K.t -> V.t -> 'a -> 'a) -> t -> 'a -> 'a
+        val bot_binop : (V.t -> V.t -> V.t) -> t -> t -> t
```
ncough: Could you access this via `meet`?

lib/analysis/wrapped_intervals.ml
agle, line 790: It doesnt make sense to assume everything is top for an intraprcedural analysis; all we know is reachable initially is the inparams to the procedure. Instead it should probably explicitly initialise the set of global variables to top in addition to the formal in params. If you want to just assume everything is top, create a new `Domain` which `include`s this and redefines `init`.

lib/analysis/rg_gen.ml
```diff
@@ -0,0 +1,392 @@
...
+
+  let transitions (lst : ConcInt.t list) =
+    Debug.print @@ "Deriving transitions from concrete interferences: "
+    ^ List.to_string ~sep:", " ConcInt.show lst;
+    (* aux derives a guarantee condition from a single (possibly simultaneous) assignment to global vars *)
+    let aux pre assignments =
+      (* true iff v := e may modify v under pre *)
+      let var_modified (v, e) =
+        (* FIXME: create fresh var with type equal to v's type *)
+        let v' = Var.create "DUMMY" Types.Integer in
```
agle, line 223: Pass through procedure and use `Procedure.fresh_var`; IF this is intended to be local (the default when no scope is passed to Var.create). Otherwise will need to pass through program's ID generator to create unique name for the global variable.

lib/analysis/rg_gen.ml
```diff
@@ -0,0 +1,392 @@
...
+        (* wrap in rvar before using in expr, i think *)
+        let v_exp = BasilExpr.rvar v in
+        let v'_exp = BasilExpr.rvar v' in
+        (* apply v' := e to get the value of v after the assignment *)
+        let assign_v' =
+          D.transfer pre
+            (Stmt.Instr_Assign { attrib = Attrib.empty; al = [ (v', e) ] })
+        in
+        (* create the expression v' != v *)
+        let v_not_eq_v' = BasilExpr.binexp ~op:`NEQ v_exp v'_exp in
```
agle, line 233: Suggest: 
```ocaml
BasilExpr.(binexp ~op:`NEQ (rvar v) (rvar v'))
```
Line by line comments restating the code are also redundant here too

lib/analysis/rg_gen.ml
```diff
@@ -0,0 +1,392 @@
...
+
+  val meet : t -> t -> t
+  (** Derives the greatest lower bound of two domain elements. *)
+
+  val havoc : t -> VarSet.t -> t
+  (** Abstracts out a set of variables; minimally weakens the domain element
+      such that the vars are not constrained. *)
+
+  val filter : t -> BasilExpr.t -> t
+  (** Returns an overapproximation (upper bound) on the states reachable after
```
agle, line 44: This sounds more like a `join` or lub with a predicate? `filter` is a slightly confusing name because its evocative of things like `List.filter`.

j-tobler: Not really a join, because "applying" the condition reduces the set of states. But I very much agree that the name should be changed to avoid clashes with ocaml idioms

lib/analysis/rg_gen.ml
j-tobler: So these are just temporary dummy variables for the purpose of doing some calculations (in this case, checking whether v := E affects v by doing DUMMY := E and then checking DUMMY == v). DUMMY is not added to the IR at all, but I was concerned about the case where v is actually called DUMMY. Since v is a global variable here, I think I need to use the program's ID generator.

lib/analysis/rg_gen.ml
j-tobler: The module is parameterised by 'rely'. Where does 'rely' come from in your suggestion?

lib/analysis/rg_gen.ml
j-tobler: This might answer the previous comment: The Analysis module is defined on line 342 over the StableD module which is defined on line 341 over the 'rely' variable, which is defined at runtime (line 376) and should change with each call to derive_guar during the fixpoint computation at line 389.

lib/analysis/rg_gen.ml
agle: Ah ok it needs to be first-class then

PR #255, opened 2026-08-28
PSTATE flag abstraction

b-paul:
This pr contains a giant ugly pattern match that identifies the semantic meaning of expressions that $PSTATE_x condition flags get assigned. A later PR will use the results to identify branch conditions hopefully! For now the transform just adds attributes to assign statements for debugging.

I also had to touch some of the algebraic simplification code to make the pattern matches slightly less horrible! This included removing a simplification that translated bvadd(a, -k) into bvsub(a, k) if k is a negative constant. I think we probably don't want to keep this as it forces analyses to handle both bvadds and bvsubs all over the place...

### DIFF
--- a/lib/lang/expr_eval.ml
+++ b/lib/lang/expr_eval.ml
@@ -88,15 +88,42 @@ let eval_expr_alg (e : Ops.AllOps.const option BasilExpr.abstract_expr) =
   | ApplyIntrin { op = #Ops.Spec.intrin } -> None
   | ApplyIntrin { op = #Ops.Maps.intrin } -> None
 
-let partial_eval_alg (e : BasilExpr.t BasilExpr.abstract_expr) :
-    BasilExpr.rewrite =
+let const_eval_expr (e : BasilExpr.t BasilExpr.abstract_expr) :
+    BasilExpr.t option =
   let open AbstractExpr in
   let open Option.Infix in
   let is_const e =
     match BasilExpr.unfix e with Constant { const } -> Some const | _ -> None
   in
-  let e = AbstractExpr.map is_const e in
-  eval_expr_alg e >|= BasilExpr.const |> BasilExpr.replace_opt
+  let e' = AbstractExpr.map is_const e in
+  eval_expr_alg e' >|= BasilExpr.const
+
+let partial_eval_intrin (e : BasilExpr.t BasilExpr.abstract_expr) :
+    BasilExpr.t option =
+  let open AbstractExpr in
+  match e with
+  | ApplyIntrin { op; args } ->
+      let consts, rest =
+        List.partition
+          (function Constant _ -> true | _ -> false)
+          (List.map BasilExpr.unfix args)
+        |> Pair.map_same (List.map BasilExpr.fix)
+      in
+      if List.length consts >= 2 then
+        let e_const =
+          BasilExpr.applyintrin ~op consts
+          |> BasilExpr.unfix |> const_eval_expr
+          |> Option.get_exn_or "terms should all be constant"
+        in
+        Some (BasilExpr.applyintrin ~op (List.append rest [ e_const ]))
+      else None
+  | _ -> None
+
+let partial_eval_alg (e : BasilExpr.t BasilExpr.abstract_expr) :
+    BasilExpr.rewrite =
+  const_eval_expr e
+  |> Option.or_lazy ~else_:(fun _ -> partial_eval_intrin e)
+  |> BasilExpr.replace_opt
 
 let partial_eval_expr e = BasilExpr.rewrite ~rw_fun:partial_eval_alg e
 let eval_expr e = BasilExpr.cata eval_expr_alg e
--- a/lib/lang/expr_eval.ml
+++ b/lib/lang/expr_eval.ml
@@ -105,9 +105,8 @@ let partial_eval_intrin (e : BasilExpr.t BasilExpr.abstract_expr) :
   | ApplyIntrin { op; args } when Ops.AllOps.is_commutative_intrin op ->
       let consts, rest =
         List.partition
-          (function Constant _ -> true | _ -> false)
-          (List.map BasilExpr.unfix args)
-        |> Pair.map_same (List.map BasilExpr.fix)
+          (function BasilExpr.E (Constant _) -> true | _ -> false)
+          args
       in
       if List.length consts >= 2 then
         let e_const =
--- a/lib/lang/expr_eval.ml
+++ b/lib/lang/expr_eval.ml
@@ -102,7 +102,7 @@ let partial_eval_intrin (e : BasilExpr.t BasilExpr.abstract_expr) :
     BasilExpr.t option =
   let open AbstractExpr in
   match e with
-  | ApplyIntrin { op; args } ->
+  | ApplyIntrin { op; args } when Ops.AllOps.is_commutative_intrin op ->
       let consts, rest =
         List.partition
           (function Constant _ -> true | _ -> false)
--- a/lib/transforms/branch_conditions.ml
+++ b/lib/transforms/branch_conditions.ml
@@ -17,13 +17,14 @@ module FlagSemantics = struct
         (** Carry when subtracting first from second *)
     | CarryBySum of Expr.BasilExpr.t * Expr.BasilExpr.t
         (** Carry when adding first and second *)
-    | NegEqual of Expr.BasilExpr.t * Expr.BasilExpr.t
+    | ZeroNegEqual of Expr.BasilExpr.t * Expr.BasilExpr.t
         (** When left == -right (or -left == right) *)
-    | Equal of Expr.BasilExpr.t * Expr.BasilExpr.t  (** When left == right *)
-    | IsZero of Expr.BasilExpr.t  (** When expr is zero *)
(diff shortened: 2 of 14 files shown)


lib/transforms/branch_conditions.ml
```diff
@@ -0,0 +1,423 @@
...
+        (** Carry when subtracting first from second *)
+    | CarryBySum of Expr.BasilExpr.t * Expr.BasilExpr.t
+        (** Carry when adding first and second *)
+    | NegEqual of Expr.BasilExpr.t * Expr.BasilExpr.t
+        (** When left == -right (or -left == right) *)
+    | Equal of Expr.BasilExpr.t * Expr.BasilExpr.t  (** When left == right *)
+    | IsZero of Expr.BasilExpr.t  (** When expr is zero *)
+    | Slt of Expr.BasilExpr.t * Expr.BasilExpr.t
+        (** When left < right signed *)
+    | IsNeg of Expr.BasilExpr.t  (** When expr is negative *)
```
katrinafyi: uniform names for the Z-based flags

lib/transforms/branch_conditions.ml
```diff
@@ -0,0 +1,423 @@
+open Bincaml_util.Common
+open Lang
+
+module FlagSemantics = struct
+  type t =
+    | Never
+    | Always
+    | OverflowByConst of Expr.BasilExpr.t * Bitvec.t
```
katrinafyi: maybe split into two ADTs, like `type flag = Z of computation and computation = Const | Diff ...`

lib/util/bitvec.ml
```diff
@@ -188,3 +188,9 @@ let repeat_bits ~(copies : int) a =
 (* extend sign bit by [extension] bits *)
 let sign_extend ~(extension : int) b =
   create ~size:(b.w + extension) @@ to_signed_bigint b
+
+(* -(2^(size)) *)
+let min_value_signed size = shl (one ~size) (of_int ~size (size - 1))
+
+(* 2^(size-1) - 1 *)
+let max_value_signed size = sub (min_value_signed size) (one ~size)
```
katrinafyi: could be `not max_value_signed`, right? but totally optional

lib/util/bitvec.ml
```diff
@@ -188,3 +188,9 @@ let repeat_bits ~(copies : int) a =
 (* extend sign bit by [extension] bits *)
 let sign_extend ~(extension : int) b =
   create ~size:(b.w + extension) @@ to_signed_bigint b
+
+(* -(2^(size)) *)
```
katrinafyi: double asterisk for odoc comment :)

lib/transforms/branch_conditions.ml
```diff
@@ -0,0 +1,423 @@
...
+  let open Stmt in
+  match stmt with
+  | Instr_Assign { attrib; al } ->
+      let annotations =
+        List.filter_map
+          (fun (v, e) ->
+            let o = FlagSemantics.extract_semantics e in
+            if
+              Option.is_none o
+              && String.starts_with ~prefix:"$PSTATE_" (Var.name v)
```
katrinafyi, line 219: rather than doing this `String.starts_with` check here, you could hoist it to a separate  `List.filter` beforehand

lib/transforms/branch_conditions.ml
```diff
@@ -0,0 +1,423 @@
...
+      ->
+        Some Always
+    | UnaryExpr
+        {
+          op = `BVNOT;
+          arg =
+            UnaryExpr
+              { op = `BOOLTOBV1; arg = BinaryExpr { op = `EQ; arg1; arg2 } };
+        } -> (
+        match (unfix3 arg1, unfix3 arg2) with
```
katrinafyi: all these cases are really big and I haven't looked closely at them. 

is it possible to split it into multiple match statements so it's less monolithic?  but I would totally understand if it's not possible.

katrinafyi (commented):
just recounting what we talked about in person so I don't forget :) the core of the PR is very good. once these small comments are done, it should be a very quick approve.

ask Allie abt alg simp/partial eval changes.

 we (me + Eris) have also talked about future testing with opcodes from llvm-mc and using the frontend (separate to this PR)

lib/transforms/branch_conditions.ml
b-paul: this check is for a debug print so terms shouldn't be filtered by it. i thought that a debug at that location made the most sense since the expression is lost after the filter_map, do you think that's fine?

lib/lang/expr_eval.ml
```diff
@@ -88,15 +88,42 @@ let eval_expr_alg (e : Ops.AllOps.const option BasilExpr.abstract_expr) =
...
-  let e = AbstractExpr.map is_const e in
-  eval_expr_alg e >|= BasilExpr.const |> BasilExpr.replace_opt
+  let e' = AbstractExpr.map is_const e in
+  eval_expr_alg e' >|= BasilExpr.const
+
+let partial_eval_intrin (e : BasilExpr.t BasilExpr.abstract_expr) :
+    BasilExpr.t option =
+  let open AbstractExpr in
+  match e with
+  | ApplyIntrin { op; args } ->
```
ncough: Probably want to check `op` is  commutative. At present this will reorder `BVConcat` arguments.

lib/transforms/branch_conditions.ml
```diff
@@ -48,178 +202,40 @@ module FlagSemantics = struct
...
         {
           op = `Extract (e1, e2);
-          arg = BinaryExpr { op = `BVSUB; arg1 = a; arg2 = b };
+          arg = BinaryExpr { op = `BVSUB; arg1 = a; arg2 = b } as arg;
         }
-      when e1 = e2 + 1 ->
+      when e1 = e2 + 1
+           && Option.equal ( = )
+                (Types.bit_width @@ type_of (fix2 arg))
+                (Some e1) ->
```
ncough: Should I read this as the intention to translate this predicate into the inverse of an `slt`? Isn't this the painful case, where you need to also check for overflow? Or is the intention to fix this up later?

lib/transforms/branch_conditions.ml
b-paul: I've broken some of the matches into different functions and simplified the checks a little bit but it is still kinda massive... unfortunately I don't think there would be a way to simplify it further :(

lib/transforms/branch_conditions.ml
b-paul: I've also removed ZNegEqual, ZEqual and NSlt cases because their validity confuses me (e.g. `R0 - 1 <= 0` is not equivalent to `R0 <= 1` because of overflow !! very evil)

lib/transforms/branch_conditions.ml
katrinafyi: Oh I was thinking hoist the var name check but keep the extract_semantics is_none check here. The idea is to separate "none because it's not a flag" and "none because extract_semantics is missing a case".

lib/lang/expr_eval.ml
```diff
@@ -88,15 +88,42 @@ let eval_expr_alg (e : Ops.AllOps.const option BasilExpr.abstract_expr) =
...
+let partial_eval_intrin (e : BasilExpr.t BasilExpr.abstract_expr) :
+    BasilExpr.t option =
+  let open AbstractExpr in
+  match e with
+  | ApplyIntrin { op; args } when Ops.AllOps.is_commutative_intrin op ->
+      let consts, rest =
+        List.partition
+          (function Constant _ -> true | _ -> false)
+          (List.map BasilExpr.unfix args)
+        |> Pair.map_same (List.map BasilExpr.fix)
```
katrinafyi: nitpicking: I found the unfix-partition-map_same kinda confusing. Could you just add an E to the predicate pattern match? up to you

lib/transforms/branch_conditions.ml
```diff
@@ -0,0 +1,363 @@
...
+    | NSlt of Expr.BasilExpr.t * Expr.BasilExpr.t
+        (** When left < right signed *)
+    | NIsNeg of Expr.BasilExpr.t  (** When expr is negative *)
+  [@@deriving show { with_path = false }]
+
+  let extract_overflow_cary arg1 arg2 =
+    let open Types in
+    let open Expr.AbstractExpr in
+    let open Expr.BasilExpr in
+    let equal e1 e2 = equal (drop_attrib e1) (drop_attrib e2) in
```
katrinafyi: equal is too generic, needs more specific name. this and other local functions can be brought into a private struct at the top of this file if you want.

katrinafyi (commented):
lgtm on my side. I'll leave the green button to Allie. thanks also to Nick.

lib/transforms/branch_conditions.ml
katrinafyi: totally fine, thanks!

lib/transforms/branch_conditions.ml
b-paul: ah sure, i was hesitant to do that since checking things based on names feels wrong but it would probably be a good idea to at least filter to variables that are bv1

katrinafyi (approved):
"epic".

PR #257, opened 2026-08-31
 consider spec in `remove-unused-decls` pass 

agle:
This pass is part of the `ssa` batch so it is annoying (and wrong) to delete shared global declarations (i.e. memory) when they're not used in program body but there exist stub procedures which implicitly modify memory.

It would be nice to make the interpretation of absent modsets in the input be "this procedure modifies nothing" but that can be a problem for later (basil doesn't emit modsets).

### DIFF
--- a/lib/transforms/ssa.ml
+++ b/lib/transforms/ssa.ml
@@ -56,20 +56,26 @@ let intro_ssi_assigns proc (should_lift : Var.t -> bool) =
   in
   Procedure.map_blocks_nondet fix_block proc
 
+(** Delete unused local var declarations and return used global variables.
+    assumes captures/modifies are up to date. *)
 let drop_unused_var_declarations_proc p =
+  let spec = Procedure.specification p in
+  let used = Procedure.free_vars_specification spec |> VarSet.of_iter in
   let used =
     Procedure.fold_blocks_topo_fwd
       (fun acc id bl ->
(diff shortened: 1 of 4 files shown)


lib/transforms/ssa.ml
```diff
@@ -56,20 +56,26 @@ let intro_ssi_assigns proc (should_lift : Var.t -> bool) =
...
+      used p
   in
   Var.Decls.filter_map_inplace
     (fun _ v -> if VarSet.mem v used then Some v else None)
     (Procedure.local_decls p);
   VarSet.filter Var.is_global used
 
+(** Update modsets and delete unused variable definitions *)
 let drop_unused_var_declarations_prog (p : Program.t) =
+  let p = Spec_modifies.set_modsets p in
```
b-paul, line 78: Could passes.ml be updated to reflect that this transform touches modsets too? And maybe changing the function name as well

b-paul (approved):
Looks good! It's a bit worrying that spec vars weren't being considered...

lib/transforms/ssa.ml
mira-alford: It may be better to have a pass that updates the modsets and this has it as a prereq. loading the ir already calls set_modsets after all so its just duplicating the pass if the modsets are not invalidated.

PR #258, opened 2026-08-31
PSTATE tracking abstract domain

b-paul:
This adds an abstract domain that tracks current flag information about PSTATE_x variables at each code point. If a flag records some information about a variable that gets updated, the abstract domain drops the information about the variable (tracking nothing instead).

This isn't needed with SSA but I thought making things work without SSA might be preferred regardless given how early in the hypothetical pipeline this transform probably would sit.

### DIFF
--- a/lib/analysis/lattice_collections.ml
+++ b/lib/analysis/lattice_collections.ml
@@ -73,7 +73,9 @@ module LatticeMap (K : MapKey) (V : TopLattice) = struct
       let name = V.name ^ "MapLattice"
       let bottom = BotMap KM.empty
       let top = TopMap KM.empty
-      let singleton k v = BotMap (KM.singleton k v)
+
+      let singleton k v =
+        if V.equal V.bottom v then bottom else BotMap (KM.singleton k v)
 
       let bot_v_op f _ x y =
         let x = f x y in
--- a/lib/analysis/lattice_collections.ml
+++ b/lib/analysis/lattice_collections.ml
@@ -196,8 +196,20 @@ module LatticeMap (K : MapKey) (V : TopLattice) = struct
         | TopMap m -> (`Top, KM.to_list m)
 
       let mapi f = function
-        | BotMap m -> BotMap (KM.mapi f m)
-        | TopMap m -> TopMap (KM.mapi f m)
+        | BotMap m ->
+            BotMap
+              (KM.filter_map
+                 (fun k v ->
+                   let v = f k v in
+                   if V.equal v V.bottom then None else Some v)
+                 m)
+        | TopMap m ->
+            TopMap
+              (KM.filter_map
+                 (fun k v ->
+                   let v = f k v in
+                   if V.equal v V.top then None else Some v)
+                 m)
 
       let fold f m acc =
         match m with BotMap m -> KM.fold f m acc | TopMap m -> KM.fold f m acc
--- a/lib/transforms/branch_conditions.ml
+++ b/lib/transforms/branch_conditions.ml
@@ -291,15 +291,12 @@ module FlagDomain = struct
         List.fold_left
           (fun m (v, e) ->
             let m = drop_modified v m in
-            (* If v is a bv1, update map with extract results *)
-            if Types.equal (Var.typ v) (Types.Bitvector 1) then
-              let f =
-                FlagSemantics.extract_semantics e
-                |> Option.map (fun f -> FlagLattice.V f)
-                |> Option.get_or ~default:FlagLattice.top
-              in
-              update v f m
-            else m)
+            let f =
+              FlagSemantics.extract_semantics e
+              |> Option.map (fun f -> FlagLattice.V f)
+              |> Option.get_or ~default:FlagLattice.top
+            in
+            update v f m)
           m al
     | _ -> m
 
--- a/lib/transforms/branch_conditions.ml
+++ b/lib/transforms/branch_conditions.ml
@@ -390,11 +390,6 @@ proc @main() -> ()
      $PSTATE_N:bv1 := extract(32,31, bvadd(bvadd(extract(32,0, $R0), bvnot(bvshl(extract(32,0, $R1), zero_extend(20, 0x0:bv12)))), 0x1:bv32));
      $PSTATE_N:bv1 := extract(32,31, bvadd(bvadd(extract(32,0, $R0), 0xffffffff:bv32), 0x1:bv32));
 
-     // should not be annotated!
-     $PSTATE_N:bv1 := extract(31,30, bvadd(extract(32,0, $R0), 0x1:bv32));
-     $PSTATE_N:bv1 := extract(31,30, bvadd(bvadd(extract(32,0, $R0), bvnot(bvshl(extract(32,0, $R1), zero_extend(20, 0x0:bv12)))), 0x1:bv32));
-     $PSTATE_N:bv1 := extract(31,30, bvadd(bvadd(extract(32,0, $R0), 0xffffffff:bv32), 0x1:bv32));
-
      $PSTATE_V:bv1 := 0x0:bv1;
      $PSTATE_C:bv1 := 0x0:bv1;
      $PSTATE_Z:bv1 := 0x1:bv1;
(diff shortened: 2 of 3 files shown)


lib/analysis/lattice_collections.ml
```diff
@@ -196,8 +196,20 @@ module LatticeMap (K : MapKey) (V : TopLattice) = struct
...
+                 (fun k v ->
+                   let v = f k v in
+                   if V.equal v V.bottom then None else Some v)
+                 m)
+        | TopMap m ->
+            TopMap
+              (KM.filter_map
+                 (fun k v ->
+                   let v = f k v in
+                   if V.equal v V.top then None else Some v)
```
katrinafyi, line 213: this seems to be doing some canonicalisation of TopMap/BotMap which isn't done in other places. maybe turn it into a standalone function like "normalise/canonicalise" so the user can manually call it after any operation if needed?

lib/transforms/branch_conditions.ml
```diff
@@ -241,7 +252,112 @@ let annotate_flag_assign_stmts (p : Program.proc) =
     (fun (bid, block) -> Block.map ~phi:id annotate_flag_assigns block)
     p
 
-let transform = annotate_flag_assign_stmts
+module L = struct
```
katrinafyi: I don't love single letter modules which aren't functor parameters

lib/transforms/branch_conditions.ml
```diff
@@ -241,7 +252,112 @@ let annotate_flag_assign_stmts (p : Program.proc) =
...
+    | Bot -> false
+    | V f -> FlagSemantics.contains_var v f
+end
+
+(** Assigns flag meaning values to flag variables at each code point. If a flag
+    assumes a value of a variable that gets updated, that flags value will get
+    dropped and set to top. This should be precise enough still as branches
+    probably only occur direct after flags are set (probably). Note that none of
+    this is a problem if ssa is run prior to this transform. *)
+module FlagAnalysis = struct
```
katrinafyi: if it makes sense, you could consider turning branch_conditions into a folder and these modules into separate files. I think it helps a lot when skimming and browsing the docs

lib/transforms/branch_conditions.ml
```diff
@@ -241,7 +252,112 @@ let annotate_flag_assign_stmts (p : Program.proc) =
...
+      Instr_Assume { attrib; body; branch }
+  | _ -> stmt
+
+(** Add flag annotations to assume statements based on flag variables prior *)
+let annotate_assume_flags (p : Program.proc) =
+  let a = A.analyse p in
+  Procedure.map_blocks_nondet
+    (fun (bid, b) ->
+      let r =
+        A.A.M.find_opt (Procedure.Vert.Begin bid) a
```
katrinafyi: 🧐 A.A.M is quite the chain of single letter modules..

lib/transforms/branch_conditions.ml
```diff
@@ -241,7 +252,112 @@ let annotate_flag_assign_stmts (p : Program.proc) =
...
+    mapi (fun v' x -> if L.contains_var v x then L.top else x)
+
+  let transfer m stmt =
+    match stmt with
+    | Stmt.Instr_Assign { al } ->
+        List.fold_left
+          (fun m (v, e) ->
+            let m = drop_modified v m in
+            (* If v is a bv1, update map with extract results *)
+            if Types.equal (Var.typ v) (Types.Bitvector 1) then
```
katrinafyi: I don't love repeating this bv1 check.

maybe extract_semantics could return a `flag result option`, where the outer option represents whether it is a flag, and the inner result records any errors during semantics extraction for a variable which *is* a flag? if that makes sense

lib/transforms/branch_conditions.ml
```diff
@@ -343,3 +459,91 @@ prog entry @main;
...
+      modifies $PSTATE_C:bv1, $PSTATE_N:bv1, $PSTATE_V:bv1, $PSTATE_Z:bv1, $R0:bv64
+      captures $PSTATE_C:bv1, $PSTATE_N:bv1, $PSTATE_V:bv1, $PSTATE_Z:bv1, $R0:bv64
+
+    [
+       block %main [
+         $PSTATE_V:bv1 := bvnot(booltobv1(eq(sign_extend(32, bvadd(extract(32,0, $R0), 0x1:bv32)), bvadd(sign_extend(32, extract(32,0, $R0)), 0x1:bv64))));
+         $PSTATE_C:bv1 := bvnot(booltobv1(eq(zero_extend(32, bvadd(extract(32,0, $R0), 0x1:bv32)), bvadd(zero_extend(32, extract(32,0, $R0)), 0x1:bv64))));
+         $PSTATE_Z:bv1 := booltobv1(eq(bvadd(extract(32,0, $R0), 0x1:bv32), 0x0:bv32));
+         $PSTATE_N:bv1 := extract(32,31, bvadd(extract(32,0, $R0), 0x1:bv32));
+         assume booland(eq($PSTATE_N, $PSTATE_V), eq($PSTATE_Z, 0x0:bv1)) { .flag_semantics_$PSTATE_C = "(O (Sum (extract(32,0, $R0), 0x1:bv32)))"; .flag_semantics_$PSTATE_N = "(N (Sum (extract(32,0, $R0), 0x1:bv32)))"; .flag_semantics_$PSTATE_V = "(O (Sum (extract(32,0, $R0), 0x1:bv32)))"; .flag_semantics_$PSTATE_Z = "(Z (Sum (extract(32,0, $R0), 0x1:bv32)))" };
```
katrinafyi, line 551: hmm can we split these expect tests into multiple smaller let%expect blocks? it would be good to clearly name which tests are expecting clobbering and which are not.

 it should also be fine to have most tests only use one flag variable to make it more concise.

katrinafyi (commented):
looks reasonable! and it would be nice to not depend on SSA. 

most of my comments are just about code organisation and hopefully it all is reasonable. let me know what you think!

lib/analysis/lattice_collections.ml
b-paul: I think (it was a while ago) my goal with this api was to make it so that the user can't construct a BotMap/TopMap with a bottom/top value. Every other function in the public interface (except singleton which im fixing) should be maintaining this invariant

Do you think this is fine?

katrinafyi: Yeah, that sounds better, in fact, but it will be a bigger change - I think we should do it in a separate PR.

In that case, we can leave the changes in this PR as-is.

b-paul: I'm confused... what would need to be changed?

lib/transforms/branch_conditions.ml
b-paul: two of these are from the intra_analysis code :sob: i've changed the outermost one to a better name though

lib/analysis/lattice_collections.ml
katrinafyi: For some reason I thought that there were lots of other functions that didn't maintain this invariant, but you're right it seems to be fine. Nothing else needed, then 🙂‍↕️ Sorry!

lib/transforms/branch_conditions.ml
katrinafyi: Appreciated! And yeah, you can leave the other ones.

lib/analysis/lattice_collections.ml
b-paul: meow

lib/transforms/branch_conditions.ml
b-paul: On reflection the check is entirely redundant since extract_semantics already always returns None if the expr isn't a bv1 (just by cases) so i've just removed it.

lib/transforms/branch_conditions.ml
b-paul: that does sound like it would help! i'll make that change in my next pr though since i've already started working off of this branch and merging everything after splitting files sounds painful

katrinafyi (approved):

PR #263, opened 2026-09-03
Branch condition expr rewrites

b-paul:
Expressions in terms of PSTATE_x variables get replaced by numerical conditions in the cases that the analysis from the previous pr can handle (except for overflow because it's ugly)! The cases that fail are ccmp instructions (as flags end up being dependent on control flow, so not constant) and when registers are overwritten after a compare (fixed by ssa i think). What's left to do other than handling these fail cases is inserting branch conditions after `PC := if c then a1 else a2` assignments (and more testing probably). yay !

I would like to split the contents of `branch_conditions.ml` into multiple files but I'd expect that doing so now would make this PR much harder to review... I'm thinking maybe splitting files could be a separate pr or just a task done after review ?


lib/transforms/branch_conditions.ml
```diff
@@ -1,10 +1,24 @@
...
+
+open struct
+  let equiv_exp e1 e2 = Expr.BasilExpr.(equal (drop_attrib e1) (drop_attrib e2))
+end
+
 module FlagSemantics = struct
   type t =
-    | Never
-    | Always
+    | Const of computation
```
katrinafyi, line 29: could be `Const of bool` right?

lib/transforms/branch_conditions.ml
```diff
@@ -14,24 +28,33 @@ module FlagSemantics = struct
     | Sum of Expr.BasilExpr.t * Expr.BasilExpr.t  (** Computed e1 + e2 *)
     | Diff of Expr.BasilExpr.t * Expr.BasilExpr.t  (** Computed e1 - e2 *)
     | Expr of Expr.BasilExpr.t  (** The result of evaluating an expr *)
+    | Always
+    | Never
   [@@deriving eq, ord, show { with_path = false }]
```
katrinafyi: these types aren't actually mutually recursive, so it's easier to read if `computation` is just put first instead of in an `and`.

lib/transforms/branch_conditions.ml
```diff
@@ -259,6 +282,18 @@ module FlagLattice = struct
     let name = "flag"
   end)
 
+  module E = Lang.Expr.BasilExpr
+
```
katrinafyi: unused?
```suggestion
```

lib/transforms/branch_conditions.ml
```diff
@@ -316,6 +351,145 @@ module FlagAnalysis = struct
...
+  (** A type of condition as described in
+      https://support.arm.com/documentation/dui0231/b/arm-instruction-reference/conditional-execution
+
+      We track one computation for each flag read, noting that sometimes not all
+      flags will be computed in the same way. *)
+  type t =
+    | EQ of FlagSemantics.computation
+    | CS of FlagSemantics.computation
+    | MI of FlagSemantics.computation
+    | VS of FlagSemantics.computation
```
katrinafyi: even in these ones, would it be clearer to include the flag name as a single-field record?
```ocaml
| EQ of { z : FlagSemantics.computation }
```

lib/transforms/branch_conditions.ml
```diff
@@ -316,6 +351,145 @@ module FlagAnalysis = struct
   let analyse p = analyse p
 end
 
+module Eval = Analysis.Intra_analysis.EvalExpr (FlagLattice)
+
+(** Rewrites boolean exprs in terms of flags to be in terms of numerical
+    conditions. *)
+module Rewriter = struct
+  (** A type of condition as described in
+      https://support.arm.com/documentation/dui0231/b/arm-instruction-reference/conditional-execution
```
katrinafyi: this link should be more official https://support.arm.com/documentation/ddi0487/mc/-Part-C-The-AArch64-Instruction-Set/-Chapter-C1-The-A64-Instruction-Set/-C1-2-Structure-of-the-A64-assembler-language/-C1-2-4-Condition-code?lang=en 
(idk what the RealView Developer Kit is, but it seems kinda old)

lib/transforms/branch_conditions.ml
```diff
@@ -316,6 +351,145 @@ module FlagAnalysis = struct
...
+        (* evaluate arg1 and arg2, if they are of the right form keep *)
+        let arg1 = Eval.eval (flip FlagDomain.read m) arg1 in
+        let arg2 = Eval.eval (flip FlagDomain.read m) arg2 in
+        match (arg1, arg2) with
+        | V (Z c), V (Const Always) -> EQ c
+        | V (C c), V (Const Always) -> CS c
+        | V (N c), V (Const Always) -> MI c
+        | V (O c), V (Const Always) -> VS c
+        | V (N c), V (O c' | Const c') -> GE { n = c; v = c' }
+        | _ -> Top)
```
katrinafyi: why is there `| Const c'` in the GE case?

also, is the reason you have used `O` to avoid clashing with this `V` from FlatLattice? If so, I would prefer to keep the flag variant as `V` so it matches the spec, and use `FlatLattice.get_val` to cast the FlatLattice V to a Some

lib/transforms/branch_conditions.ml
```diff
@@ -316,6 +351,145 @@ module FlagAnalysis = struct
...
+        | Some (E (BinaryExpr { op = `BVULE; arg1; arg2 })) ->
+            Some (binexp ~op:`BVULT arg2 arg1)
+        | Some (E (BinaryExpr { op = `BVULT; arg1; arg2 })) ->
+            Some (binexp ~op:`BVULE arg2 arg1)
+        | Some (E (BinaryExpr { op = `BVSLE; arg1; arg2 })) ->
+            Some (binexp ~op:`BVSLT arg2 arg1)
+        | Some (E (BinaryExpr { op = `BVSLT; arg1; arg2 })) ->
+            Some (binexp ~op:`BVSLE arg2 arg1)
+        | Some e -> Some (Expr.BasilExpr.boolnot e)
+        | None -> None)
```
katrinafyi, line 488: could maybe be handled by a smart constructor. maybe in future work.

lib/transforms/branch_conditions.ml
```diff
@@ -653,9 +837,63 @@ prog entry @main;
...
+      modifies $PSTATE_C:bv1, $PSTATE_N:bv1, $PSTATE_V:bv1, $PSTATE_Z:bv1
+      captures $PSTATE_C:bv1, $PSTATE_N:bv1, $PSTATE_V:bv1, $PSTATE_Z:bv1, $R0:bv64
+
+    [
+       block %main [
+         $PSTATE_V:bv1 := bvnot(booltobv1(eq(sign_extend(64, bvadd(bvadd($R0, 0xfffffffffffffffe:bv64), 0x1:bv64)), bvadd(bvadd(sign_extend(64, $R0), 0xfffffffffffffffffffffffffffffffe:bv128), 0x1:bv128))));
+         $PSTATE_C:bv1 := bvnot(booltobv1(eq(zero_extend(64, bvadd(bvadd($R0, 0xfffffffffffffffe:bv64), 0x1:bv64)), bvadd(bvadd(zero_extend(64, $R0), 0xfffffffffffffffe:bv128), 0x1:bv128))));
+         $PSTATE_Z:bv1 := booltobv1(eq(bvadd(bvadd($R0, 0xfffffffffffffffe:bv64), 0x1:bv64), 0x0:bv64));
+         $PSTATE_N:bv1 := extract(64,63, bvadd(bvadd($R0, 0xfffffffffffffffe:bv64), 0x1:bv64));
+         assume bvslt(0x1:bv64, $R0);
```
katrinafyi, line 904: really cool!!!

lib/util/recursionscheme.ml
```diff
@@ -195,8 +195,8 @@ module Recursion (O : Fix) = struct
     let alg r x = (alg1 r x, alg2 (map_expr snd x)) in
     fst (fst % map_fold ~f ~alg r, snd % map_fold ~f ~alg r) e
 
-  (** catamorphism that also passes the original expression map_exprped through
-      a function argument *)
+  (** catamorphism that also passes the original expression map_expred through a
+      function argument *)
```
katrinafyi: is this meant to mean like `map_expr`-ed? past-tense of `map_expr`?? seems confusing

katrinafyi (commented):
lgtm! small comments :)

lib/transforms/branch_conditions.ml
b-paul: hmm this is a little annoying... the branch conditions type works over computations and sometimes needs to handle cases when flags are always/never (e.g. cmp x0, #0 then b.gt will have overflow be never) so storing consts as bools would mean a conversion would have to be performed ... im not sure if i prefer that over this though it is unsatisfying that Const _ could represent non constant things

lib/transforms/branch_conditions.ml
b-paul: The const c' case comes from something like `cmp r0, #0` then `b.ge ...` where the overflow flag simplifies to false but other flags don't. Other than with `b.gt` i haven't seen any cases where some flags are nonconstant but others are and hopefully with further testing (aslp lifting of random instructions) if such cases exist they'll be spotted

there's no reason i used O instead of V, i think i was just thinking overflow when i wrote the code the first time but i agree with you that matching the spec makes more sense (it seems not to clash with the flatlattice V ! thank you ocaml)

lib/transforms/branch_conditions.ml
katrinafyi: I seee. We can leave it for now.

lib/transforms/branch_conditions.ml
katrinafyi: Hmm, okay! I may ask you more questions about the const variant in person.

Thanks for renaming O :)

katrinafyi (approved):
thankss

PR #268, opened 2026-09-07
split branch reconstruction stuff into folders

b-paul:
huzzah


lib/transforms/branch_conditions/cfg_analysis.ml
```diff
@@ -0,0 +1,77 @@
...
+        rhs
+        |> List.map (fun (_, k) -> read k m)
+        |> List.fold_left FlagLattice.join FlagLattice.bottom
+        |> fun v -> drop_modified lhs m |> update lhs v
+end
+
+module FlagAnalysis = struct
+  include Analysis.Intra_analysis.Forwards (FlagDomain)
+
+  let analyse p = analyse p
```
katrinafyi: which `analyse` does this reference? is it redundant?

lib/transforms/branch_conditions/flags.ml
```diff
@@ -0,0 +1,217 @@
...
+  | C of computation  (** Carry from computation *)
+  | Z of computation  (** When computation is zero *)
+  | N of computation  (** When computation is negative *)
+[@@deriving eq, ord, show { with_path = false }]
+
+let equiv_computations c c' =
+  let open Expr.BasilExpr in
+  match (c, c') with
+  | Sum (e1, e2), Sum (e1', e2') -> equiv_exp e1 e1' && equiv_exp e2 e2'
+  | Diff (e1, e2), Diff (e1', e2') -> equiv_exp e1 e1' && equiv_exp e2 e2'
```
katrinafyi: could fold these 2 cases using |

lib/transforms/branch_conditions/flags.ml
```diff
@@ -0,0 +1,217 @@
...
+  | Expr e, Expr e' -> equiv_exp e e'
+  | _ -> false
```
katrinafyi: missing Always and Never cases? maybe better to avoid `_` in this pattern match.

katrinafyi (commented):
awesome!! quick comments which I noticed while reading, let me know what you think

lib/transforms/branch_conditions/flags.ml
b-paul: without the _ there would be 20 extra cases to handle when the patterns are different :(

katrinafyi: hm that's true. what about explicitly matching c and use `_` for c'? so you only need N patterns at most, and this will still get an exhaustive warning when it changes.


```suggestion
  | (Sum _ | Diff _ | Expr _ | Always | Never), _ -> false
```

katrinafyi (approved):

PR #269, opened 2026-09-07
branch condition reconstruction: giant aslp cram test

b-paul:
as the title says

i went through most of the instructions in <https://support.arm.com/documentation/ddi0602/2024-03/Base-Instructions> that seemed to touch the control flags and added them to this big il file! note that some instructions are aliased from the perspective of control flow (e.g. adds and cmn are the same, ands and tst are the same). The one instruction that didn't work was adcs (add with carry set flags) which might need a little bit of work im not sure yet


katrinafyi (approved):
great, thanks!

PR #279, opened 2026-09-11
PC if then else transform

b-paul:
This PR adds a transform that inserts blocks with guards after a `PC = if c then addr1 else addr2` assign! See the doc comment in pc_ite.ml for an example. The `.address` attribute of a block is used so maybe an invariant should be added to the pass...

### DIFF
--- a/lib/transforms/branch_conditions/pc_ite.ml
+++ b/lib/transforms/branch_conditions/pc_ite.ml
@@ -1,0 +1,109 @@
+(** Rewrite gotos after PC if-then-else (ite) assignments to assert conditions
+    on the ite before the target goto block.
+
+    {[
+    block %block_22 [
+      $PC:bv64 := if boolnot(eq($R19, 0x0:bv64)) then 0xce14:bv64 else 0xce08:bv64;
+      goto (%paclist_get_code_21);
+    ];
+    block %paclist_get_code_21 [
+      assert boolor(eq(0xce08:bv64, $PC), eq(0xce14:bv64, $PC));
+      goto (%paclist_get_code_14,%paclist_get_code_1);
+    ];
+    ]}
+    Here, before [%paclist_get_code_14] we would want to assume either
+    [boolnot(eq($R19, 0x0:bv64))] or its negation! This module performs such
+    rewrites. *)
+
+open Lang
+open Common
+
+module PcValue = struct
+  type 'a t = { cond : 'a; t_case : Bitvec.t; f_case : Bitvec.t }
+  [@@deriving show { with_path = false }, ord, eq]
+
+  let extract_value (f : Expr.BasilExpr.t -> 'a option) (e : Expr.BasilExpr.t) :
+      'a t option =
+    let open Expr.BasilExpr in
+    match unfix3 e with
+    | ApplyIntrin
+        {
+          op = `Cases;
+          args =
+            [
+              BinaryExpr
+                {
+                  op = `IfThen;
+                  arg1 = cond;
+                  arg2 = Constant { const = `Bitvector t_case };
+                };
+              Constant { const = `Bitvector f_case };
+            ];
+        } ->
+        f (fix cond) |> Option.map (fun cond -> { cond; t_case; f_case })
+    | _ -> None
+end
+
+(** A (semi-direct?) product domain that simultaneously finds flags and uses
+    those flags to identify PC values. *)
+module PcDomain = struct
+  open Cfg_analysis
+
+  let name = "TODO"
+
+  type t = {
+    flags : FlagDomain.t;
+        [@printer fun fmt -> fprintf fmt "%s" % FlagDomain.show]
+    pc : Rewriter.t PcValue.t option;
+  }
+  [@@deriving show { with_path = false }, ord, eq]
+
+  let pretty = Containers_pp.text % show
+  let top = { flags = FlagDomain.top; pc = None }
+  let bottom = { flags = FlagDomain.bottom; pc = None }
+
+  let join a b =
+    {
+      flags = FlagDomain.join a.flags b.flags;
+      pc =
+        (if Option.equal (PcValue.equal Rewriter.equal) a.pc b.pc then a.pc
+         else None);
+    }
+
+  let leq a b = FlagDomain.leq a.flags b.flags
+  let widening = join
+  let narrowing = const
+
+  let init ?vertex _ =
+    match vertex with Some (Some Procedure.Vert.Entry) -> top | _ -> bottom
+
+  let transfer state stmt =
+    match stmt with
+    | Stmt.Instr_Assign { al } ->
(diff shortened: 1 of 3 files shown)


b-paul:
i've added a test but accidentally --amend ed the last commit instead of adding a new one oops

lib/transforms/branch_conditions/pc_ite.ml
```diff
@@ -0,0 +1,166 @@
...
+      goto (%paclist_get_code_21);
+    ];
+    block %paclist_get_code_21 [
+      assert boolor(eq(0xce08:bv64, $PC), eq(0xce14:bv64, $PC));
+      goto (%paclist_get_code_14,%paclist_get_code_1);
+    ];
+    ]}
+    Here, before [%paclist_get_code_14] we would want to assume either
+    [boolnot(eq($R19, 0x0:bv64))] or its negation! This module performs such
+    rewrites. *)
```
katrinafyi: yay for documentation! i think this comment could be clearer if it included (at least the beginning of) the `%paclist_get_code_14` block which will have the PC address so then you wouldn't have to say "or" in  "XXX or its negation!"

also extra things:
- could change the blockk names to be more illustrative rather than directly from the example
- could include a simple ascii diagram of the structure............................. maybe yoo much tho

lib/transforms/branch_conditions/pc_ite.ml
```diff
@@ -0,0 +1,166 @@
...
+end
+
+module PcValue = struct
+  type 'a t = { cond : 'a; t_case : Z.t; f_case : Z.t }
+  [@@deriving show { with_path = false }, ord, eq]
+
+  let extract_value (f : Expr.BasilExpr.t -> 'a option) (e : Expr.BasilExpr.t) :
+      'a t option =
+    let open Expr.BasilExpr in
+    match unfix3 e with
```
katrinafyi: comment about the purpose of this? i think it's to filter the condition of the ITE expression in some way? 

also, why the need for try/catch? is it expected/observed that some expressions overflow?

lib/transforms/branch_conditions/pc_ite.ml
```diff
@@ -0,0 +1,166 @@
...
+          let f_case = Bitvec.value f_case in
+          f (fix cond) |> Option.map (fun cond -> { cond; t_case; f_case })
+        with Z.Overflow -> None)
+    | _ -> None
+end
+
+module PcDomain = struct
+  open Cfg_analysis
+
+  let name = "TODO"
```
katrinafyi: ? :)

lib/transforms/branch_conditions/pc_ite.ml
```diff
@@ -0,0 +1,166 @@
...
+  let open Option.Infix in
+  let* pc = match a with Pc p -> Some p | _ -> None in
+  let cond = pc.cond in
+  let ncond = Expr.BasilExpr.unexp ~op:`BoolNOT pc.cond in
+  let* l_addr = Attrib.find_int_map ".address" l.attrib in
+  let* r_addr = Attrib.find_int_map ".address" r.attrib in
+  if
+    (not @@ Z.equal l_addr r_addr)
+    && List.mem ~eq:Z.equal pc.t_case [ l_addr; r_addr ]
+    && List.mem ~eq:Z.equal pc.f_case [ l_addr; r_addr ]
```
katrinafyi: is this just set equals???? it is very good to allow the changed order. however, it is inscrutable at first glance.

consider:
- CCList.subset in both directions? https://c-cube.github.io/ocaml-containers/containers/CCList/index.html#val-subset
- using a Set? but we'd have to make a Set functor if it's not alread made.

lib/transforms/branch_conditions/pc_ite.ml
```diff
@@ -0,0 +1,166 @@
...
+    in
+    let p = Procedure.modify_succs p bid ~remove:[ lid; rid ] ~add:[ tb; fb ] in
+    let p =
+      if Z.equal pc.t_case l_addr then
+        Procedure.modify_succs p tb ~remove:[] ~add:[ lid ] |> fun p ->
+        Procedure.modify_succs p fb ~remove:[] ~add:[ rid ]
+      else
+        Procedure.modify_succs p tb ~remove:[] ~add:[ rid ] |> fun p ->
+        Procedure.modify_succs p fb ~remove:[] ~add:[ lid ]
+    in
```
katrinafyi: ok so this is more accounting for switched order?? can we somehow "fix" the order in an earlier step and then later steps don't have to worry about it?

at least a commet :)

test/transforms/test_branch_conditions.ml
```diff
@@ -362,3 +362,57 @@ prog entry @main;
...
+       block %main [ $PC:bv64 := if boolnot(eq($R0, 0x0:bv64)) then 0xce14:bv64 else 0xce08:bv64; goto (%b1); ];
+       block %b1 [ assert boolor(eq(0xce08:bv64, $PC), eq(0xce14:bv64, $PC)); goto (%block_1,%block); ];
+       block %block [ guard boolnot(eq($R0, 0x0:bv64)); goto (%b2); ];
+       block %b2 { .address = 52756 } [ assume eq(0xce14:bv64, $PC); goto (%ret); ];
+       block %block_1 [ guard boolnot(boolnot(eq($R0, 0x0:bv64))); goto (%b3); ];
+       block %b3 { .address = 52744 } [ assume eq(0xce08:bv64, $PC); goto (%ret); ];
+       block %ret [ return; ]
+    ];
+    prog entry @main;
+    |}]
```
katrinafyi: for my peace of mind, can we have a test case that works on an ITE structure where the two destinations coincide? e.g., arising from `b.eq #4`.

lib/transforms/branch_conditions/pc_ite.ml
```diff
@@ -0,0 +1,166 @@
...
+    rewrites. *)
+
+open Lang
+open Common
+
+open struct
+  let equiv_exp e1 e2 = Expr.BasilExpr.(equal (drop_attrib e1) (drop_attrib e2))
+end
+
+module PcValue = struct
```
katrinafyi: not sure about the name PcValue, since idk.. it's not just one "PC value" since there's branches and it actually represents an ITE including the condition.

just a thought but doesn't need to change.

katrinafyi (approved):
thanks!! it looks pretty cool and is pretty neat. 

i'll have a poke at improving the switched order code

lib/transforms/branch_conditions/pc_ite.ml
katrinafyi: I made a commit here to try and improve this logic, if you want to use it: https://github.com/UQ-PAC/bincaml/commit/2e053b1c8d5d307249c913e788ac99c55a402d97

